PR #110, opened 2015-01-06
Fix for RESTWS-482

akmad:


omod-common/src/main/java/org/openmrs/module/webservices/rest/util/ReflectionUtil.java
```diff
@@ -17,7 +17,13 @@
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
+import com.sun.tools.javac.util.Pair;
```
djazayeri: We should not include anything in com.sun.(*)

omod-common/src/main/java/org/openmrs/module/webservices/rest/util/ReflectionUtil.java
```diff
@@ -25,6 +31,25 @@
  */
 public class ReflectionUtil {
 	
+	private static ConcurrentMap<Pair<DelegatingResourceHandler<?>, String>, Method> setterMethodCache;
```
djazayeri: Offhand I think we should clear these caches when the module activator's contextRefreshed is called.

omod/src/main/java/org/openmrs/module/webservices/rest/Activator.java
```diff
@@ -34,6 +37,11 @@ public void started() {
 	
 	@Override
 	public void stopped() {
+		log.info("Clearing caches...");
+		
+		ConversionUtil.clearCache();
```
djazayeri: per my earlier comment, I think this should happen on contextRefreshed...

omod-common/src/main/java/org/openmrs/module/webservices/rest/util/ReflectionUtil.java
akmad: Is there any Pair-like class that I can use here? I had previously added a dependency to javatuples but was told to remove that as well. I know it's not a big deal and I can just concat the two into a String or something, but this will be the second time that a fix has done this in the restws module; at some point doesn't it just make sense to add a dependency which provides this?

djazayeri: I would have just added a CacheKey class.
But I don't have a big problem importing javatuples, if it's a very small library...

PR #111, opened 2015-01-22
RESTWS-487 - Fixing update of userProperties on UserResource

shruthidipali:


djazayeri:
In the ticket you said "old value for userProperties is returned as a map. unchangedValue does not handle the case of the property being a map."

It seems like here you are just skipping whatever is supposed to be done if a map is passed in. Is that right? Shouldn't we change unchangedValue to handle maps?

PR #113, opened 2015-03-05
Restws 488

seekme94:
The required feature was to set _startDatetime_ property for _Visit_ to current date if not provided by the client.


rkorytkowski:
@mhawila, thanks for the review! @seekme94, well done!

PR #114, opened 2015-03-09
RESTWS-493 - adding user info to session

namratanehete:


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/SessionController.java
```diff
@@ -43,7 +45,17 @@
...
 	@ResponseBody
 	public Object get(WebRequest request) {
-		return new SimpleObject().add("sessionId", request.getSessionId()).add("authenticated", Context.isAuthenticated());
+		boolean authenticated = Context.isAuthenticated();
+		SimpleObject session = new SimpleObject();
+		session.add("sessionId", request.getSessionId()).add("authenticated", authenticated);
+		if (authenticated) {
+			SimpleObject authenticatedUser = new SimpleObject();
+			User user = Context.getAuthenticatedUser();
+			authenticatedUser.add("uuid", user.getUuid()).add("display",
```
djazayeri: Instead of manually building a SimpleObject here, you should do:

    ConversionUtil.convertToRepresentation(Context.getAuthenticatedUser(), Representation.DEFAULT)

namratanehete: I tried to do that first, but the converter for User.class is in the omod-18, where as the sessioncontroller is in omod-common. If omod-common depends on omod-18 it will be cyclic dependency and all modules, other than common depend on both omod-18 and omod-common.

djazayeri: Can we move sessioncontroller into omod-18?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_8/SessionController1_8.java
```diff
@@ -43,7 +49,24 @@
...
 	@ResponseBody
 	public Object get(WebRequest request) {
-		return new SimpleObject().add("sessionId", request.getSessionId()).add("authenticated", Context.isAuthenticated());
+            boolean authenticated = Context.isAuthenticated();
+            SimpleObject session = new SimpleObject();
+            session.add("sessionId", request.getSessionId()).add("authenticated", authenticated);
+            if (authenticated) {
+                User user = Context.getAuthenticatedUser();
+                String rep = request.getParameter(RestConstants.REQUEST_PROPERTY_FOR_REPRESENTATION);
+                if(null == rep || RestConstants.REPRESENTATION_DEFAULT.equals(rep)){
```
djazayeri: We never want to manually construct a simple object. We always want to return a proper representation of the User.

Pseudo code for this:

    Representation rep = getRepresentation(request); // DEFAULT if nothing is specified. And use RestService.getRepresentation(String) for this
    def userRep = ConversionUtil.convertToRepresentation(Context.getAuthenticatedUser(), rep);

PR #115, opened 2015-03-15
RESTWS-456:Cannot unretire locations (or other metadata) via REST

lluismf:
https://issues.openmrs.org/browse/RESTWS-456


djazayeri:
Need changes specific to metadata vs data. See ticket.

PR #118, opened 2015-04-24
RESTWS-499 "Encounter creation fails for Obs mapping with ConceptMapping...

tomaszmueller:
... instead of Concept uuid"


tomaszmueller:
@rkorytkowski, @wluyima https://issues.openmrs.org/browse/RESTWS-499

rkorytkowski:
@tomaszmueller please write a test case to prove it actually works

tomaszmueller:
@rkorytkowski updated and created tests cases

PR #122, opened 2015-06-17
Preethi, #RESTWS-507 | Named representation is not picked up correctl…

preethi29:
…y when there are more than one named representation in a resource


sunbiz:
A unit test that fails before this fix is applied would be good to have as part of this ticket. @preethi29 , can you please add a unit test on any existing Resource that has a problem to this PR?

preethi29:
@sunbiz Have added a test for ConceptResource, please check

dkayiwa:
@sunbiz do you wanna give this a second review? :)

sunbiz:
Thanks @dkayiwa for the ping... @preethi29 merged this. Thanks for the fix.

PR #123, opened 2015-06-20
RESTWS-509 Added SystemSettingsResource to get Global Properties

namratanehete:


omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/SystemSettingResource1_9.java
```diff
@@ -0,0 +1,255 @@
...
+/**
+ * {@link Resource} for {@link GlobalProperty}, supporting standard CRUD
+ * operations
+ */
+@Resource(name = RestConstants.VERSION_1 + "/systemsetting", supportedClass = GlobalProperty.class, supportedOpenmrsVersions = {"1.9.*", "1.10.*", "1.11.*"})
+public class SystemSettingResource1_9 extends DelegatingCrudResource<GlobalProperty> {
+
+    public static final String GENERAL = "General Settings";
+
+    @RepHandler(RefRepresentation.class)
```
djazayeri: By convention the ref representation should only have uuid, display, and links. Also, you are missing the default representation, which should look more like this.

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/SystemSettingResource1_9.java
```diff
@@ -0,0 +1,255 @@
...
+        description.addProperty("handlerConfig");
+        description.addSelfLink();
+        return description;
+    }
+
+    /**
+     * @see DelegatingCrudResource#getRepresentationDescription(Representation)
+     */
+    @Override
+    public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
```
djazayeri: Why use RepHandler for two of the representations, and getRepresentationDescription for the third? I would suggest you just do if (ref) ... else if (default) ... else if (full) ... in this method

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/SystemSettingResource1_9.java
```diff
@@ -0,0 +1,255 @@
...
+        String section = GENERAL;
+        String name = globalProperty.getProperty();
+        int sectionEnd = globalProperty.getProperty().indexOf(".");
+        if (sectionEnd > 0) {
+            section = globalProperty.getProperty().substring(0, sectionEnd);
+            section = beautify(section);
+            name = globalProperty.getProperty().substring(sectionEnd + 1);
+            name = beautify(name);
+        }
+        return section + " - " + name + " - " + globalProperty.getPropertyValue();
```
djazayeri: I think that the second " - " should be an =, e.g. name = value.

omod-1.9/src/test/java/org/openmrs/module/webservices/rest/web/RestTestConstants1_9.java
```diff
@@ -1,44 +1,46 @@
-package org.openmrs.module.webservices.rest.web;
-
-public class RestTestConstants1_9 {
```
djazayeri: Something went wrong, probably with formatting / line spacing. What actually changed here, and is it possible to avoid modifying the entire file to add a line?

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/SystemSettingResource1_9.java
```diff
@@ -0,0 +1,255 @@
...
+    public static final String GENERAL = "General Settings";
+
+    @RepHandler(RefRepresentation.class)
+    public SimpleObject asRef(GlobalProperty delegate) throws ConversionException {
+        DelegatingResourceDescription description = new DelegatingResourceDescription();
+        description.addProperty("uuid");
+        description.addProperty("property");
+        description.addProperty("propertyValue");
+        description.addProperty("description");
+        description.addProperty("display", findMethod("getDisplayName"));
```
djazayeri: Please don't use "findMethod" but rather give that method a @PropertyGetter annotation.

See https://issues.openmrs.org/browse/RESTWS-502

djazayeri:
There is one overall issue that needs to be changed: starting with OpenMRS 1.9, the GlobalProperty object supports custom datatypes, handlers etc, but you're not properly using those here.

The idea is that if the admin has set a datatypeClassname, then the framework automatically knows how to convert the propertyValue between String, and some other type. You should be exposing the "value" property as the primary one, and only exposing "propertyValue" as a secondary thing.

See https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/BaseAttributeCrudResource1_9.java for an example of how to handle the "value" property.

namratanehete:
Creating a new PR with the fixes

PR #136, opened 2015-08-11
RESTWS-448: added optional search parameters for date from and to to …

bashra:
…encounter resource

### DIFF
--- a/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_8/EncounterSearchHandler1_8.java
+++ b/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_8/EncounterSearchHandler1_8.java
@@ -19,12 +19,11 @@ import java.util.Collections;
 import java.util.Date;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_8/EncounterSearchHandler1_8.java
```diff
@@ -51,13 +58,20 @@ public SearchConfig getSearchConfig() {
     public PageableResult search(RequestContext context) throws ResponseException {
         String patientUuid = context.getRequest().getParameter("patient");
         String encounterTypeUuid = context.getRequest().getParameter("encounterType");
+        
+        String dateFrom = context.getRequest().getParameter(DATE_FROM);
+        String dateTo = context.getRequest().getParameter(DATE_TO);
+        
+        Date fromDate = dateFrom!=null ? isoParser.parseDateTime(dateFrom).toDate() : null;
```
wluyima: Use ConversionUtil.convert otherwise this might not work for other supported date formats, you wouldn't also need to include the changes in the pom file.

PR #137, opened 2015-08-12
RESTWS-520: Ability to remove the last selected program state

shanmugam-radhakrishnan:


shanmugam-radhakrishnan:
It looks like the code is not readable so instead of adding comment to the code, I have done some code refactor. Hope it clear now.

PR #143, opened 2015-08-20
RESTWS-521: Adding Audit info to PatientStateResource

SwathiVarkala:
…ientSTateResource


dkayiwa:
@SwathiVarkala did you take a look at the auditInfo property?

dkayiwa:
@SwathiVarkala auditinfo is not for all representations.

SwathiVarkala:
@dkayiwa I have added auditInfo in PatientState REF representation also because if i add it only in FULL and when i request for PatientEnrollmentResource Full representation i get recursively the PatientState, PatientProgram data which means lot of duplicate data. So to avoid that I am querying for PatientEnrollmentRource Default Representation which has PatientState Ref representation and added AuditInfo in it. As Default representation will contain all the info in ref had to add AuditInfo in Default also.

dkayiwa:
You will need to first discuss this on talk to get what others feel about returning auditInfo in the default representation.

PR #152, opened 2015-09-06
RESTWS-523: Added a FormResource sub-resource to Form resource

mhawila:
Along with the FormResource sub-resource, I have added controllers to handle ClobDatatypeStorage upload/download/delete.


mhawila:
@dkayiwa is there anything I would need to address in these changes?

dkayiwa:
@mhawila did you address the clobdata rename issue?

mhawila:
@dkayiwa I think I explained earlier that I clobdata is meant to be generic and I if one refers to form resource the they will have to use the form/{uuid}/formresource/{uuid}/value which internally calls clobdata controller but if someone wants to upload clob data unrelated to form resource they will be able to do so without any naming confusion.

dkayiwa:
@mhawila so will one have to use formresource/clobdata instead of simply formresource/formresourcedata?

mhawila:
@dkayiwa No, if one wants to go via form resource the url is .../ws/rest/v1/form/{uuid}/formresource/{resourceUuid}/value. On the other hand if one wants to deal with other non form resource clob data the uri is .../ws/rest/v1/clobdata

PR #153, opened 2015-09-08
RESTWS-529: Fixed incorrect behaviour when submitting uuid for coded …

mhawila:
This is a fix to the way the Property setter for value handles coded boolean values. Initially the passed was being compared to the Concept object for true/false (i.e Yes/No). That implementation caused the comparison to always return false. This fix makes the comparison against the concept uuid instead which is the expected behaviour.

### DIFF
--- a/omod-1.8/pom.xml
+++ b/omod-1.8/pom.xml
@@ -11,7 +11,7 @@
 	<description>OpenMRS module project for Rest Web Services</description>
 
 	<properties>
-		<openmrs.version.1.8>1.8.1</openmrs.version.1.8>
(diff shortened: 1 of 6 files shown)


dkayiwa:
@mhawila do you mind adding a unit test for this?

mhawila:
Well I could even though I thought it is kind of straight forward :)

dkayiwa:
Even what is straight forward can be broken. Just exactly like how this got broken. :)
So the test will ensure we catch the offender red handed. :)

mhawila:
@dkayiwa I made the changes you proposed. This should be good to go now.

PR #173, opened 2015-11-05
RESTWS-421:Remove all usages of findMethod in resources (so that all properties are available in custom representations)

lluismf:
https://issues.openmrs.org/browse/RESTWS-421


lluismf:
You're right, it should have been deprecated, not removed. Can you cherry pick this commit? https://github.com/openmrs/openmrs-module-webservices.rest/commit/842b4a3c83b06a50ae89b1437b87a058f385887b

PR #181, opened 2015-12-14
RESTWS-304:added tests for ConceptSearchHandler

Werick:


Werick:
I notice its not doing anything so far, I think it should be dropped

dkayiwa:
Will you drop it? :)

Werick:
I already dropped and the pr is updated

Werick:
Yes that is the term uuid

PR #184, opened 2016-02-04
RESTWS-552: Removed unimplemented fetch all operations from docs.

psbrandt:
The Swagger documentation generator was assuming that all resources supported the fetch all operation. This commit introduces a check to ensure that some subclass of `*BaseDelegatingResource` actually implements the `doGetAll` method.


psbrandt:
The other strategy here would have been to call the `doGetAll` method and see in an exception was shown, but I thought this approach might be more efficient.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -94,6 +93,16 @@ private void CreateApiDefinition() {
 		swaggerSpecification.setConsumes(consumes);
 	}
 	
+	private boolean TestOperationImplemented(OperationEnum operation, BaseDelegatingResource resource) {
+		Method method = null;
+		switch (operation) {
+			case get:
+				method = ReflectionUtils.findMethod(resource.getClass(), Operation.OPERATION_GET_ALL_METHOD, RequestContext.class);
```
wluyima: I doubt if this is guaranteed to work because a resource can implement a method and just throw an exception, this is why i would assume invoking the method and catching the ResourceDoesNotSupportOperationException is still the better option

psbrandt:
@wluyima There seems to be one case of us returning `null` [here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ActiveListTypeResource1_8.java#L70). I guess that's a bug?

psbrandt:
Fixed the above mentioned bug in b2e651d.

psbrandt:
@wluyima I implemented a first attempt at your suggestion. Let me know what you think when you have time.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -94,6 +89,33 @@ private void CreateApiDefinition() {
...
+				rc.setLimit(1);
+				
+		}
+		
+		return rc;
+	}
+	
+	private boolean TestOperationImplemented(OperationEnum operation, BaseDelegatingResource resource) {
+		switch (operation) {
+			case get:
```
wluyima: i would imagine you would do this for all operations because can chose not to support any operation

wluyima:
Looks good, i added a comment about other operations, not only getAll

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
psbrandt: The scope of the ticket I created (RESTWS-552) is just the fetch all operation, and I created this PR for that ticket, but I'll the others to this PR if you think I should.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -94,6 +89,33 @@ private void CreateApiDefinition() {
 		swaggerSpecification.setConsumes(consumes);
 	}
 	
+	private RequestContext BuildDummyRequestContextForOperationTest(OperationEnum operation) {
+		RequestContext rc = new RequestContext();
+		rc.setRequest(((ServletRequestAttributes) RequestContextHolder.getRequestAttributes()).getRequest());
```
wluyima: You can use an instance of spring's MockHttpServletRequest here

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
wluyima: ok

psbrandt:
Sorry, please excuse the backup commit. I'll squash it and let you guys know when this PR is ready for another review.

wluyima:
Ok

psbrandt:
I'm doing a few more things than my branch name suggest, so I'll close this and open a new PR when I'm done.

PR #186, opened 2016-02-09
Fix for: RESTWS-419 - WSD-DATE date parsing not quite ISO8601 compatible

bholagabbar:
**Issue**: [RESTWS-419](https://issues.openmrs.org/browse/RESTWS-419)

**Fix**: Replace the the non ISO8601 `SimpleDateFormat` in Java with the `DateTime` class parser which is a part of the [Joda-Time] (http://www.joda.org/joda-time/) public library. This parses ISO compliant dates correctly and fixes the issue where TimeZones were being dropped.
A relevant Unit Test has also been added

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
@@ -37,6 +37,8 @@ import java.util.concurrent.ConcurrentMap;
 import org.apache.commons.beanutils.PropertyUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.joda.time.DateTime;
+import org.joda.time.format.DateTimeFormat;
 import org.openmrs.Auditable;
 import org.openmrs.Retireable;
 import org.openmrs.Voidable;
@@ -238,15 +240,15 @@ public class ConversionUtil {
 
 
 			if (toClass.isAssignableFrom(Date.class)) {
-				ParseException pex = null;
+				IllegalArgumentException pex = null;
 				String[] supportedFormats = {"yyyy-MM-dd'T'HH:mm:ss.SSSZ", "yyyy-MM-dd'T'HH:mm:ss.SSS",
 						"yyyy-MM-dd'T'HH:mm:ssZ", "yyyy-MM-dd'T'HH:mm:ssXXX", "yyyy-MM-dd'T'HH:mm:ss", "yyyy-MM-dd HH:mm:ss", "yyyy-MM-dd"};
 				for (int i = 0; i < supportedFormats.length; i++) {
 					try {
-						Date date = new SimpleDateFormat(supportedFormats[i]).parse(string);
+						Date date = DateTime.parse(string, DateTimeFormat.forPattern(supportedFormats[i])).toDate();
 						return date;
 					}
-					catch (ParseException ex) {
+					catch (IllegalArgumentException ex) {
 						pex = ex;
 					}
 				}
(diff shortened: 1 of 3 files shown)


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
```diff
@@ -238,15 +240,15 @@ public static Object convert(Object object, Type toType) throws ConversionExcept
...
 
 			if (toClass.isAssignableFrom(Date.class)) {
-				ParseException pex = null;
+				IllegalArgumentException pex = null;
 				String[] supportedFormats = {"yyyy-MM-dd'T'HH:mm:ss.SSSZ", "yyyy-MM-dd'T'HH:mm:ss.SSS",
 						"yyyy-MM-dd'T'HH:mm:ssZ", "yyyy-MM-dd'T'HH:mm:ssXXX", "yyyy-MM-dd'T'HH:mm:ss", "yyyy-MM-dd HH:mm:ss", "yyyy-MM-dd"};
 				for (int i = 0; i < supportedFormats.length; i++) {
 					try {
-						Date date = new SimpleDateFormat(supportedFormats[i]).parse(string);
+						Date date = DateTime.parse(string, DateTimeFormat.forPattern(supportedFormats[i])).toDate();
```
lluismf: DateTime.parse(String str) is not smart enough to detect all supported formats?

bholagabbar: I tried to maintain the flow as it was. Should I change it? Also, incase of wrong input, it throws an `IllegalArgumentException`

lluismf: No, it's just an improvement that could be done. 
About the exception, if it happens the client won't know the reason but that's a general problem of the module.

bholagabbar: Right so I think we can leave it this way. If you see the diff, all I did was change the parser. Rest is as it is

djazayeri: If you're importing Joda time, shouldn't you be able to just use their ISODateTimeFormat directly and get rid of "supportedFormats"?

bholagabbar: Yeah I tried that. Apparently, 3 tests fail when I do
 `Date date = ISODateTimeFormat.dateTime().parseDateTime(string).toDate();`  There errors are:

1. _convert_shouldReturnFalseOnIncorrectDateFromString()_ throws error `java.lang.IllegalArgumentException: Invalid format: "2012-05-01T00:00:00.000" is too short`

2. _convert_shouldReturnCorrectDateWhenParsingStringHavingTimeZone()_ throws error `java.lang.IllegalArgumentException: Invalid format: "2016-01-12T06:00:00+05:30" is malformed at "+05:30"`

3. _convert_shouldReturnEqualsDateFromString()_ throws error `java.lang.IllegalArgumentException: Invalid format: "2011-05-01" is too short`
<br>
Hence I decided to the accepted formats given in the string and parse them the same way it was being done before. @djazayeri

bholagabbar: @lluismf As for `Date date = DateTime.parse(string).toDate();`, the last test, _convert_shouldReturnEqualsDateFromString()_ fails. The error thrown is `org.openmrs.module.webservices.rest.web.response.ConversionException: Error converting date - correct format (ISO8601 Long): yyyy-MM-dd'T'HH:mm:ss.SSSZ`
<br>
Taking the workflow and everything related to the tests in consideration, I felt that the way I have currently implemented the parsing seems to work out fine :)

PR #189, opened 2016-02-24
RESTWS-496 Handle internal errors gracefully

jdegraft:


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
```diff
@@ -91,7 +107,15 @@ public Object create(SimpleObject propertiesToCreate, RequestContext context) th
 		
 		T delegate = handler.newDelegate();
 		setConvertedProperties(delegate, propertiesToCreate, handler.getCreatableProperties(), true);
-		ValidateUtil.validate(delegate);
+		
+		try {
```
wluyima: This is not the thing that needs to be addressed based on the comments @lluismf added, in any case this is taken care of by BaseRestController.validationExceptionHandler.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
```diff
@@ -91,7 +107,15 @@ public Object create(SimpleObject propertiesToCreate, RequestContext context) th
 		
 		T delegate = handler.newDelegate();
 		setConvertedProperties(delegate, propertiesToCreate, handler.getCreatableProperties(), true);
-		ValidateUtil.validate(delegate);
+		
+		try {
+			ValidateUtil.validate(delegate);
```
lluismf: I think save(delegate) already invokes the validator, so no need to invoke it twice.

jdegraft: ValidateUtil.validate was already being called, I only wrapped it in a try catch block to return the errors to the webservices client. In other words, it seems it was already being called twice.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
jdegraft: If webservices is already returning error messages as I recall from earlier work in adminui where I had to write a general angular handler to process errors from webservices then what is this ticket supposed to do.
I don't seem to understand the requirement.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
wluyima: save(delegate) doesn't invoke the validator, the original code was correct and this try catch isn't necessary because wrapping of the errors and sending an http error code is already handled by the framework, see the code in BaseRestController.validationExceptionHandler

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
wluyima: You actually just answered yourself and confirmed what i was trying say, the webservices module's framework already catches validation exceptions, wraps and sends back as part of the response with the appropriate error code as seen [here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/BaseRestController.java#L80),  in the adminUI module all you did was to detect the same error code and the messages send by this logic in BaseRestController.validationExceptionHandler, does this make sense now?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/BaseRestController.java
```diff
@@ -83,6 +84,18 @@ public SimpleObject validationExceptionHandler(ValidationException validationExc
...
 	}
 	
+	/**
+	 * Handle ConvertionException - return response with exception message - ConversionUtil throws
+	 * ConversionException
+	 */
+	@ExceptionHandler(ConversionException.class)
+	@ResponseBody
+	public SimpleObject conversionExceptionHandler(ConversionException conversionException, HttpServletRequest request,
+	        HttpServletResponse response) {
```
wluyima: This seems to address the issue, there is existing code that did this except that it didn't send back the error message, may be it was intentional not to do so since error messages can contain sensitive data you might want to sent back to any user?

dkayiwa: @jdegraft did you see the above comment?

jdegraft: updated pull request

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/RestUtil.java
```diff
@@ -807,9 +807,9 @@ public static boolean hasCause(Throwable throwable, Class<? extends Throwable> c
 	public static SimpleObject wrapErrorResponse(Exception ex, String reason) {
 		LinkedHashMap map = new LinkedHashMap();
 		if (reason != null && !reason.isEmpty()) {
-			map.put("message", reason);
+			map.put("message", reason + "[" +  ex.getMessage() + "]");
```
dkayiwa: Don't you want to add some space between reason and the rest?

jdegraft:
updated pull request

lluismf:
:joy:

PR #191, opened 2016-03-07
📗 Swagger improvements.

psbrandt:
This a refactor of the Swagger specification creation code. An attempt has been made to make the code more readable/maintainable and bring the `json` we build *mostly* up to [spec](http://swagger.io/specification/).

The following issues have also been specifically addressed:

 - [x] **RESTWS-554**: Module resources not added to Swagger docs
 - [x] **RESTWS-553**: DELETE operations not documented by Swagger
 - [x] **RESTWS-552**: Docs generated by Swagger in cases where operation not supported
 - [x] **RESTWS-543**: Get /visit is not displayed in swagger documentation

Lastly, I've published the source used to build the bundled Swagger UI along with some documentation on how to contribute [here](https://github.com/psbrandt/openmrs-contrib-apidocs/tree/swagger-ui).


rkorytkowski:
Thanks Pascal! Could you please send separate pull requests for issues in the future (or at least separate commits)? It's really hard to review such a big change. I'll pick up your branch for testing today and get back to you if any fixes are needed.

psbrandt:
That was definitely the intention @rkorytkowski, but I ended up doing a pretty significant refactoring of [the old SwaggerDocumentationCreator](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java) to [the new one](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java). As a result of this refactor, most of the existing issues just disappeared.

I'll try to summarise the changes that apply to each ticket here.

### RESTWS-543: Get /visit is not displayed

The problem was that because empty delegates were being used to generate the docs, sometimes the converters were falling over because of NPEs. This was fixed by just adding a few more null checks [here](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java#L97). 

Strictly, these checks may no longer be necessary, since I've actually completely removed the need to use delegates and migrated to just using `getRepresentation` instead. This simplified things a lot, but isn't perfect, because some resources use the `@RepHandler` annotation, which I don't currently make use of.

### RESTWS-552: Unsupported operations documented

A lot of assumptions were being made about what operations are actually supported for each resource. To fix this, I wrote the [`testOperationImplemented`](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L176) method. This method checks to see if the `ResourceDoesNotSupportOperationException` is thrown for a specific operation, and if so, does not document the operation.

### RESTWS-553: DELETE operations not documented

The `DELETE` and `PURGE` operations simply weren't implemented, so I did that [here](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L822).

### RESTWS-554: Module resources not added

This was an issue with how the resource names were being generated, which I improved [here](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L767).

### Other Improvements

There are still many improvements that could be made, but I think I made the code more readable and maintainable by simplifying some of the complicated control flow and breaking down the code into smaller methods.

#### Noisy Logs

I added a global property and a [`toggleLogs`](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L96) method that silences logging while the docs are being generated. This isn't ideal, but in my opinion it's better than getting hundreds of exceptions dumped in the logs while we test each operation on each resource with junk data. Anyway, it's configurable, so people who want to see the exceptions can do so.

#### Unused Code

I removed a bunch of classes that were related to the old way of generating the docs. I'm still making use of some of this code (e.g. `ResourceRepresentation`), so it hasn't all been taken out.

#### Small Fixes

There are one or two resources that I fixed that either had small typos (e.g. an extra `/` in an annotation attribute) or returned `null` instead of throwing `ResourceDoesNotSupportOperationException`.

### Testing

Since we are now testing to check that resources support specific operations, we need to make sure that calling the methods on specific resources does not result in persistent changes. I wrote [this test](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java), which checks database row counts before and after generating the docs.

This isn't perfect, however, since not all possible resources are present when generating the docs inside this unit test. I'm happy to hear suggestions on how we can improve testing.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/Tag.java
```diff
@@ -54,4 +54,8 @@ public void setDescription(String description) {
 		this.description = description;
 	}
 	
+	@Override
+	public int compareTo(Tag tag) {
+		return name.compareToIgnoreCase(tag.getName());
```
wluyima: Don't you need to check if tag is not null?

omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.GlobalProperty;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.webservices.docs.swagger.SwaggerSpecificationCreator;
+import org.openmrs.module.webservices.rest.web.RestConstants;
+import org.openmrs.module.webservices.rest.web.api.RestService;
+import org.openmrs.web.test.BaseModuleWebContextSensitiveTest;
+
+import java.sql.*;
```
wluyima: We don't use wildcard imports by convention

omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java
```diff
@@ -0,0 +1,120 @@
...
+	
+	Map<String, Integer> beforeCounts;
+	
+	public Map<String, Integer> getRowCounts() throws Exception {
+		Map<String, Integer> ret = new HashMap<String, Integer>();
+		
+		Connection con = this.getConnection();
+		DatabaseMetaData metaData = con.getMetaData();
+		
+		ResultSet rs = metaData.getTables(null, "PUBLIC", "%", null);
```
wluyima: Won't you end up with the same results with these arguments as null?

omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java
```diff
@@ -0,0 +1,120 @@
...
+		while (rs.next()) {
+			String tableName = rs.getString(3);
```
wluyima: You can get table row count from the DatabaseConnection object:

`new DatabaseConnection(getConnection()).getRowCount(tableName)`

omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java
```diff
@@ -0,0 +1,120 @@
...
+			
+			ret.put(tableName, resultSet.getInt("rowcount"));
+			
+			resultSet.close();
+		}
+		
+		return ret;
+	}
+	
+	private void showTableContents(String tableName) throws Exception {
```
wluyima: Core already has a similar method that prints table contents:

`TestUtils.printOutTableContents`

omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java
```diff
@@ -0,0 +1,120 @@
...
+		SwaggerSpecificationCreator ssc = new SwaggerSpecificationCreator("/v1/");
+		ssc.BuildJSON();
+		
+		Map<String, Integer> afterCounts = getRowCounts();
+		
+		Assert.assertTrue(ensureCountsEqual(beforeCounts, afterCounts));
+	}
+	
+	private boolean ensureCountsEqual(Map<String, Integer> beforeCounts, Map<String, Integer> afterCounts) throws Exception {
+		if (beforeCounts.size() != afterCounts.size()) {
```
wluyima: You could use junit assertions to check if equal that takes a message too to be printed in case the assertion fails

wluyima:
I actually can't see the diff for SwaggerSpecificationCreator because it is too large, i think the you could have committed the swagger files separately since they are part of your actual code changes

psbrandt:
> I actually can't see the diff for `SwaggerSpecificationCreator` because it is too large

I know, it was a pretty big refactor. What can we do about it?

> I think the you could have committed the swagger files separately since they are part of your actual code changes

You're right. Do you want me to pull them out, or just do things better next time? :)

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/Tag.java
psbrandt: From [here](https://docs.oracle.com/javase/7/docs/api/java/lang/Comparable.html):

> Note that `null` is not an instance of any class, and `e.compareTo(null)` should throw a `NullPointerException`

This is what will happen, but I could make it explicit if you prefer?

omod-common/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java
psbrandt: Nope:

```
Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 15.574 sec <<< FAILURE!
checkNoDatabaseChanges(org.openmrs.module.webservices.rest.doc.SwaggerSpecificationCreatorTest)  Time elapsed: 15.084 sec  <<< ERROR!
org.h2.jdbc.JdbcSQLException: Table "CATALOGS" not found; SQL statement:
SELECT count(*) AS rowcount FROM CATALOGS [42102-135]
```

wluyima:
Don't you need to always reset System.out and err when you're done?

psbrandt:
> Don't you need to always reset System.out and err when you're done?

I do that [here](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L131-L132) with [this](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L83) method call:

```java
toggleLogs(RestConstants.SWAGGER_LOGS_ON);
```

wluyima:
so you call toggleLogs twice i.e before and after generating the docs?

wluyima:
I think [this](https://github.com/psbrandt/openmrs-module-webservices.rest/blob/swagger-improvements/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L81) code needs to be in a try finally clause, to guarantee that the oupout streams always get reset regardless of whether the operation fails.

psbrandt:
> so you call toggleLogs twice i.e before and after generating the docs?

Yup:
```java
	public String BuildJSON() {
		synchronized (this) {
			toggleLogs(RestConstants.SWAGGER_LOGS_OFF);
			createApiDefinition();
			addPaths();
			toggleLogs(RestConstants.SWAGGER_LOGS_ON);
			addDefinitions();
			addSubclassOperations();
		}
		return createJSON();
	}
```

wluyima:
I see, did you see the comment about the try finally clause?

psbrandt:
> did you see the comment about the try finally clause?

Yeah, working on it.

psbrandt:
Can this PR be closed, since the code was merged by @wluyima?

PR #193, opened 2016-03-16
RESTWS-564: Created a setValue Property Setter to make it do the righ…

mhawila:
…t thing for Attributables

### DIFF
--- a/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonAttributeResource1_8.java
+++ b/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonAttributeResource1_8.java
@@ -34,6 +34,7 @@ import org.openmrs.module.webservices.rest.web.representation.Representation;
 import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
 import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingSubResource;
 import org.openmrs.module.webservices.rest.web.resource.impl.NeedsPaging;
+import org.openmrs.module.webservices.rest.web.response.ConversionException;
 import org.openmrs.module.webservices.rest.web.response.ResponseException;
 import org.openmrs.util.OpenmrsClassLoader;
 
@@ -76,31 +77,11 @@ public class PersonAttributeResource1_8 extends DelegatingSubResource<PersonAttr
 	@Override
 	public DelegatingResourceDescription getCreatableProperties() {
 		DelegatingResourceDescription description = new DelegatingResourceDescription();
-		description.addProperty("value");
 		description.addRequiredProperty("attributeType");
+		description.addProperty("value");
 		description.addProperty("hydratedObject");
 		return description;
 	}
-
-    /*
-        We may need to hydrate Attributables in the request. To do that we need to know
-        its type. So we can't set hydratedObject until attributeType has been set.
-        Since SimpleObject is a LinkedHashMap, it is ordered, and the following method is
-        overridden to ensure hydratedObject will appear last in the properties.
-     */
-    @Override
-    public Object create(String parentUniqueId, SimpleObject post, RequestContext context) throws ResponseException {
-        uglyMethodToEnsureHydratedObjectWillBeSetLast(post);
-        return super.create(parentUniqueId, post, context);
-    }
-
-    private void uglyMethodToEnsureHydratedObjectWillBeSetLast(SimpleObject post) {
-        Object hydratedObject = post.get("hydratedObject");
-        if (hydratedObject != null) {
-            post.remove("hydratedObject");
-            post.put("hydratedObject", hydratedObject);
-        }
-    }
 
     @PropertySetter("hydratedObject")
     public void setHydratedObject(PersonAttribute personAttribute, String attributableUuid) {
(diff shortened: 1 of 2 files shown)


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonAttributeResource1_8.java
```diff
@@ -76,32 +77,12 @@ public DelegatingResourceDescription getRepresentationDescription(Representation
...
 
-    /*
-        We may need to hydrate Attributables in the request. To do that we need to know
-        its type. So we can't set hydratedObject until attributeType has been set.
-        Since SimpleObject is a LinkedHashMap, it is ordered, and the following method is
-        overridden to ensure hydratedObject will appear last in the properties.
-     */
-    @Override
-    public Object create(String parentUniqueId, SimpleObject post, RequestContext context) throws ResponseException {
-        uglyMethodToEnsureHydratedObjectWillBeSetLast(post);
```
dkayiwa: @mhawila since you removed this, will the new setValue property setter ensure that the hydrated object is set last?

mhawila: yes it is ensured by the order properties are added in getCreatableProperties() method. And since   [DelegatingResourceDescription](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingResourceDescription.java#L33-L35) uses LinkedHashMap internally the order is guaranteed to be maintained.

lluismf: It's dangerous to rely on the internal representation of the properties.

mhawila: Why?

lluismf: https://en.wikipedia.org/wiki/Law_of_Demeter

Even if PersonAttributeResource1_8 extends DelegatingResourceDescription , the internal representation of properties should be hidden and PersonAttributeResource1_8 should not assume it's ordered.

dkayiwa: @mhawila in the method that you removed above, i see its javadoc explicitly mentioning the LinkedHashMap. But still goes ahead to do the uglyMethodToEnsureHydratedObjectWillBeSetLast method. Did this later on change in any way not to require it anymore?

lluismf: I'm wondering why SimpleObject has to be an ordered Map.

mhawila: @dkayiwa its because initially the properties were enumerated from the payload, in essence the order in the payload determined which was evaluated first but it was later changed to be enumerated based on the order in the creatable properties [See here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingResource.java#L597). Besides this method is not called if you are submitting the attribute via its containing super resource, so you will still have the problem.

mhawila: @lluismf in this case the super class does not require the sub to know that the properties are ordered. It is the sub that takes advantage of that. It is like the sub tells the super I want things in a certain order and I expect that to be honored, that sounds like a very good design principle.

lluismf: I recall a similar problem with the Obs resource, setters must be invoked in a certain order otherwise won't work. I think it's a code smell and not a good design. Because all the sets are done by reflection there's no room to improve this. Also, it's the philosophy of the REST module.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonAttributeResource1_8.java
```diff
@@ -113,6 +94,37 @@ public void setHydratedObject(PersonAttribute personAttribute, String attributab
...
 
+	@PropertySetter("value")
+	public  void setValue(PersonAttribute personAttribute, String value) {
+		PersonAttributeType attributeType = personAttribute.getAttributeType();
+		if(attributeType == null) {
+			personAttribute.setValue(value);
+		} else {
+			// Check if expected value is attributable and do the right thing.
+			try {
+				Class<?> clazz = Context.loadClass(personAttribute.getAttributeType().getFormat());
```
dkayiwa: What happens when format is NULL?

mhawila: If format is null an NPE is thrown. I guess I need to test for that.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonAttributeResource1_8.java
```diff
@@ -113,6 +94,42 @@ public void setHydratedObject(PersonAttribute personAttribute, String attributab
...
+		if(attributeType == null) {
+			personAttribute.setValue(value);
+		} else {
+			// Check if expected value is attributable and do the right thing.
+			try {
+				String format = personAttribute.getAttributeType().getFormat();
+				if(format == null) {
+					personAttribute.setValue(value);
+				} else {
+					Class<?> clazz = Context.loadClass(personAttribute.getAttributeType().getFormat());
```
dkayiwa: You could as well have used the already existing format variable instead of making the call again to get the format.

mhawila: Yeah I could, I was fast :) sorry!

PR #224, opened 2016-07-21
💲 RESTWS-579: Ensure both legacy UI and Swagger can access jQuery

psbrandt:
Previously, Swagger UI was including its own copy of jQuery (and preventing OpenMRS from including a copy). This caused code generated by the header in the legacy UI to fall over💥.

This PR changes things so that the default jQuery provided by the legacy UI is used instead. However, since Swagger UI expects the 💲 variable to exist, we also have to do the following hack because the legacy UI uses `jQuery.noConflict()`:

```js
var $ = jQuery; 
```


rkorytkowski:
Fair enough! Thanks!

PR #232, opened 2016-09-17
RESTWS-612 Merge inner wrapper composite key classes into one

teleivo:
two nested classes with almost identical implementation where used as a wrapper
class so two strings could be used as keys to the maps holding SearchHandlers.

* renamed one of them to CompositeSearchHandlerKey
* first property stays supportedResource, second property is simply
additionalKeyProperty which is once populated with the SearchHandler id and
once with the parameter names.

see https://issues.openmrs.org/browse/RESTWS-612

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
@@ -49,9 +49,9 @@ public class RestServiceImpl implements RestService {
 	
 	volatile Map<Class<?>, Resource> resourcesBySupportedClasses;
 	
-	private volatile Map<SearchHandlerParameterKey, Set<SearchHandler>> searchHandlersByParameter;
+	private volatile Map<CompositeSearchHandlerKey, Set<SearchHandler>> searchHandlersByParameter;
 	
-	private volatile Map<SearchHandlerIdKey, SearchHandler> searchHandlersByIds;
+	private volatile Map<CompositeSearchHandlerKey, SearchHandler> searchHandlersByIds;
 	
 	private volatile Map<String, Set<SearchHandler>> searchHandlersByResource;
 	
@@ -83,39 +83,59 @@ public class RestServiceImpl implements RestService {
 		
 	}
 	
-	private static class SearchHandlerParameterKey {
+	/**
+	 * Wraps {@code Resource} name and an additional string-based key into a composite key.
+	 */
+	private static class CompositeSearchHandlerKey {
 		
-		public String supportedResource;
+		public final String supportedResource;
 		
-		public String parameter;
+		public final String additionalKeyProperty;
 		
-		public SearchHandlerParameterKey(String supportedResource, String parameter) {
+		public CompositeSearchHandlerKey(String supportedResource, String additionalKeyProperty) {
 			this.supportedResource = supportedResource;
-			this.parameter = parameter;
+			this.additionalKeyProperty = additionalKeyProperty;
 		}
 		
+		/**
+		 * @see Object#hashCode()
+		 * @return the hash code
+		 * @should return same hashcode for equal composite keys
+		 */
 		@Override
 		public int hashCode() {
 			final int prime = 31;
 			int result = 1;
-			result = prime * result + ((parameter == null) ? 0 : parameter.hashCode());
+			result = prime * result + ((additionalKeyProperty == null) ? 0 : additionalKeyProperty.hashCode());
 			result = prime * result + ((supportedResource == null) ? 0 : supportedResource.hashCode());
 			return result;
 		}
 		
+		/**
+		 * @see Object#equals(Object)
+		 * @param obj the object to test for if equal to this
+		 * @return true if obj is equal to this otherwise false
+		 * @should return true if given this
+		 * @should be symmetric
+		 * @should be transitive
+		 * @should return false if given null
+		 * @should return false if given an object which is not an instanceof this class
+		 * @should return false if given a composite key with different supported resource
+		 * @should return false if given a composite key with different additional key
+		 * @should return false if given a composite key with null as additional key
+		 * @should return false if given a composite key with null as supported resource
+		 */
 		@Override
 		public boolean equals(Object obj) {
 			if (this == obj)
 				return true;
-			if (obj == null)
+			if (!(obj instanceof CompositeSearchHandlerKey))
 				return false;
-			if (getClass() != obj.getClass())
-				return false;
-			SearchHandlerParameterKey other = (SearchHandlerParameterKey) obj;
-			if (parameter == null) {
-				if (other.parameter != null)
+			CompositeSearchHandlerKey other = (CompositeSearchHandlerKey) obj;
+			if (additionalKeyProperty == null) {
+				if (other.additionalKeyProperty != null)
 					return false;
-			} else if (!parameter.equals(other.parameter))
+			} else if (!additionalKeyProperty.equals(other.additionalKeyProperty))
 				return false;
 			if (supportedResource == null) {
 				if (other.supportedResource != null)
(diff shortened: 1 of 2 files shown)


teleivo:
@dkayiwa what are your thoughts on this one? reduces duplication of inner class which just acts as a wrapper for a key composed of two strings in a Map. Are you for it/against it? Should I add tests for the class equals/hashcode for you to be for it ;)

dkayiwa:
@teleivo just add the tests and will then review for feedback. :)

teleivo:
@dkayiwa updated the PR with tests, finally a readable diff (thank you formatting :trophy: )

The only thing I have doubts and would like your input is, I made the inner class public so I can test it. Do you feel like the class should be a top level class? So it can be used elsewhere?

Should I make it private again and add some reflection magic to keep it testable?

dkayiwa:
@teleivo if you do not yet know where else it will be used, assume it will never. :)
It is generally great to designe classes in a way that makes them testable. But if the only reason for making the class more visible is such that you can test it, then i would rather use some reflection magic.

teleivo:
@dkayiwa agree to not making things public if not needed. but is it then necessary to test the private class? also arguable if that is the best practice. Tests of the public things in the RestServiceImpl should prove that it works as expected regardless of its inner implementation details which are hidden.

So why test the inner class at all? Wasnt done before that either (I know thats not the argument but the reason might have been my argument before).

I think either make it public and test it. Or make it private and not test it via reflection.

So what shall I do?

teleivo:
@dkayiwa ok, made the inner class private. tests use reflection to create an instance. please have a look :+1:

dkayiwa:
@teleivo have you looked at the cause of the coveralls failure?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -83,88 +83,57 @@ public ResourceDefinition(Resource resource, int order) {
...
-	}
-	
-	private static class SearchHandlerIdKey {
+	/**
+	 * Wraps {@code Resource} name and an additional string-based key into a composite key.
+	 */
+	private static class CompositeSearchHandlerKey {
 		
-		public String supportedResource;
+		final public String supportedResource;
```
dkayiwa: Our undocumented convention is to have public before final :)

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -83,88 +83,57 @@ public ResourceDefinition(Resource resource, int order) {
...
 		
-		public String id;
+		final public String additionalKeyProperty;
```
dkayiwa: Same here.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -83,88 +83,59 @@ public ResourceDefinition(Resource resource, int order) {
...
+	/**
+	 * Wraps {@code Resource} name and an additional string-based key into a composite key.
+	 */
+	private static class CompositeSearchHandlerKey {
 		
-		public String supportedResource;
+		public final String supportedResource;
 		
-		public String id;
+		public final String additionalKeyProperty;
```
dkayiwa: What value does "additional" add to the key variable name?

teleivo:
@dkayiwa what should i call it?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
dkayiwa: Would "key" make it less meaningful? :)

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -83,88 +83,59 @@ public ResourceDefinition(Resource resource, int order) {
...
 		
-		public SearchHandlerIdKey(String supportedResource, String id) {
+		public CompositeSearchHandlerKey(String supportedResource, String additionalKeyProperty) {
 			this.supportedResource = supportedResource;
-			this.id = id;
-		}
-		
-		public SearchHandlerIdKey(SearchHandler searchHandler) {
-			this.supportedResource = searchHandler.getSearchConfig().getSupportedResource();
```
dkayiwa: Did you just drop this constructor that take a SearchHandler parameter?

teleivo:
@dkayiwa yes dropped it, since I cannot determine what prop of the SearchHandler should go into the second key (the ID, a certain param, ...).

teleivo:
@dkayiwa so I'll rename additionalKeyProperty to key

Anything else?

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImplTest.java
```diff
@@ -268,4 +271,161 @@ public void getResourceBySupportedClass_shouldReturnTheMostExactMatch() throws E
...
+	@Test
+	public void equals_shouldReturnFalseIfGivenACompositeKeyWithNullAsAdditionalKey() throws Exception {
+		Class compositeSearchHandlerClass = Class
+		        .forName("org.openmrs.module.webservices.rest.web.api.impl.RestServiceImpl$CompositeSearchHandlerKey");
+		Constructor ctor = compositeSearchHandlerClass.getDeclaredConstructors()[0];
+		
+		Object compositeKey1 = ctor.newInstance("v1/order", "default");
+		Object compositeKey2 = ctor.newInstance(null, "default");
+		assertFalse(compositeKey1.equals(compositeKey2));
+		assertFalse(compositeKey2.equals(compositeKey1));
```
rkorytkowski: It really looks bad. I saw your discussion with @dkayiwa, but don't you both feel it turned out weird?

You really should be testing the exposed behavior and not if equals works as expected. You will gain more by testing:
```
firstSearchHandler = mock(SearchHandler.class);
firstSearchConfig = new SearchConfig("default", "v1/order", ..., ...);
when(firstSearchHandler.getSearchConfig()).thenReturn(firstSearchConfig);

secondSearchHandler = mock(SearchHandler.class);
secondSearchConfig = new SearchConfig("default", null, ..., ...);
when(secondSearchHandler.getSearchConfig()).thenReturn(secondSearchConfig);

when(restSeviceHelper.getSearchHandlers()).thenReturn(Arrays.asList(firstSearchHandler, secondSearchHandler));

expectedException.expect(InvalidSearchConfigException.class);
expectedException.expectMessage("The resource name for search config should not be null")
restService.getResourceByName("v1/order");
```
When trying to run the test you will actually discover that it fails when calling secondSearchConfig = new SearchConfig("default", null, ..., ...);, which means the violation was caught even earlier.

teleivo:
@rkorytkowski didn't want to test the private class. Agree.

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImplTest.java
teleivo: @rkorytkowski 

I get that we should test the public API of this class. I will remove the tests of the private nested class and write missing tests of the public stuff.

when trying your snippet I get a `java.lang.IllegalArgumentException: supportedResource must not be empty`

And your point here is that this shows that we are testing something that can never happen? Or that the nested class has an issue?

rkorytkowski: Yes, it can never happen, because it is prevented much earlier when using public api we expose. The test is still useful though. Simply expect IlegalArgumentException.class when passing null as supportedResource to the SearchConfig constructor.

teleivo: but now your talking about a test for the `SearchConfig` and  which by the way has no Test class at all. this is another ticket (which Ill create).

teleivo:
@rkorytkowski  @dkayiwa am closing this for now.
will make a PR adding tests to public methods of RestServiceImpl first.
then I will open this one again without tests of private methods.

PR #239, opened 2016-09-20
RESTWS-611 Use maven license plugin

teleivo:
* replace license.txt by file from openmrs-core/LICENSE
* added license-header.txt from openmrs-core
* update license headers in all files, they still had the old version 1 :exclamation: 

* use com.mycila maven license plugin
takes license-header.txt from this repo and updates .txt/.xml/.java files

* runs in root/api/omod
* runs by default plugin goal format which formats the headers if not properly
formatted or missing
* when environment variable CI=true it will it will run goal "check" which
fails the build if a header is missing or not properly formatted

try for yourself(change a header or delete one):
formatting
```
mvn license:format
```

check
```
mvn license:check
```

simulate being on travis CI
```
export CI=true
mvn clean install -DskipTests
```

simulate being a dev :bowtie: 
```
export CI=false
mvn clean install -DskipTests
```

see https://issues.openmrs.org/browse/RESTWS-611


teleivo:
@dkayiwa please only merge this if it is ok to update the license from version 1 to 2. seems that the license file in this repo was not updated.

also its bit trickier here with the profiles since it used activeByDefault on "java-8" profile.
"java-8" profile is still activated by default but through a workaround. this is necessary since otherwise it can not be combined with profile "ci" which will be activated on travis CI.

am wondering if this can be an issue with the "non-java-8" profile. so please be careful and tell me :grinning:

dkayiwa:
@teleivo you can always test out the non java8 profile. :)

teleivo:
@dkayiwa now works like this:

defaults to java 8 profile
```
mvn clean package
```

non java 8 profile
```
mvn clean package -DnoJava8
```

and "ci" profile can be added on top of above ones if env variable CI=true

:mega: :new:

teleivo:
@dkayiwa I know you said next time split formatting changes into its own PR but here its just not working. I need to update the license headers otherwise the build fails since the plugin complains about the license headers not being correct. kind of :chicken: :egg: issue here, hope you can :fork:giveme :)

PR #245, opened 2016-09-23
RESTWS-623 Make RestServiceImpl.getResourceByName testable

teleivo:
* add dependency for OpenmrsClassScanner to RestServiceImpl and inject via xml
using the factory method on OpenmrsClassScanner.getInstance() which returns the
singleton instance
* add first tests for getResourceByName by mocking the
OpenmrsClassScanner.getClasses

see https://issues.openmrs.org/browse/RESTWS-623

### DIFF
--- a/omod-common/src/test/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImplTest.java
+++ b/omod-common/src/test/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImplTest.java
@@ -9,13 +9,20 @@
  */
 package org.openmrs.module.webservices.rest.web.api.impl;
 
+import org.mockingbird.test.rest.resource.DuplicateNameAndOrderMockingBirdFantasyResource;
+import org.mockingbird.test.rest.resource.DuplicateNameMockingBirdFantasyResource;
+import org.mockingbird.test.rest.resource.MockingBirdFantasyNameResource;
+import org.mockingbird.test.rest.resource.MockingBirdFantasyResource;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import org.mockingbird.test.rest.resource.UnannotatedMockingBirdFantasyResource;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
 import org.openmrs.Patient;
 import org.openmrs.Person;
+import org.openmrs.api.APIException;
+import org.openmrs.module.webservices.rest.web.OpenmrsClassScanner;
 import org.openmrs.module.webservices.rest.web.RestUtil;
 import org.openmrs.module.webservices.rest.web.api.RestHelperService;
 import org.openmrs.module.webservices.rest.web.api.RestService;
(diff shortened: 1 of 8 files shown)


teleivo:
@rkorytkowski hope you can review this PR and tell me if the approach is sound.

some notes:
webModuleApplicationContext.xml seems to change totally in the diff. but I only added the 

```xml
	<bean id="openmrsClassScanner"
 		  class="org.openmrs.module.webservices.rest.web.OpenmrsClassScanner"
 		factory-method="getInstance" >
 	</bean>
```

and injected it into the added RestServiceImpl property.

Another thing I could use your help is:

in the RestServiceImplTest I added a fake `MockingBirdFantasyResource` so I can test (otherwise I need to add a dependency to another maven module which has Resources)

the issue I am having is that the the one test which returns the resource works only if the openmrs version is equal to the current one. I hard coded "1.9.10" since I cannot use the static `OpenmrsConstants.OPENMRS_VERSION_SHORT` maybe you have an idea for me on how to handle this :)

other than that hope we are going in the right direction :tropical_drink:

teleivo:
@dkayiwa or @rkorytkowski 

created a helper method in RestServiceImplTest to set the current openmrs version (this resolves the issue I talked about in the last comment.

remaining notes:

webModuleApplicationContext.xml seems to change totally in the diff (I think this is due to formatter not formatting xml and the line endings are now handled by git via .gitattributes which wasnt before). but I only added the

```java
    <bean id="openmrsClassScanner"
          class="org.openmrs.module.webservices.rest.web.OpenmrsClassScanner"
        factory-method="getInstance" >
    </bean>
```

and injected it into the added RestServiceImpl property.

teleivo:
@dkayiwa or @rkorytkowski

last note on this! its ready to be merged from my side.

* created a helper method in RestServiceImplTest to set the current openmrs version
* created a fake/test package org/mockingbird/test/rest/resource/ with Resource/SubResource's used in the RestServiceImplTest
(this separate package is needed otherwise these fake resources will be picked up by the OpenmrsClassScanner and influence other tests.)
* added tests

remaining notes:

webModuleApplicationContext.xml seems to change totally in the diff (I think this is due to formatter not formatting xml and the line endings are now handled by git via .gitattributes which wasnt before). but I only added the

    <bean id="openmrsClassScanner"
          class="org.openmrs.module.webservices.rest.web.OpenmrsClassScanner"
        factory-method="getInstance" >
    </bean>
and injected it into the added RestServiceImpl property.

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImplTest.java
```diff
@@ -147,6 +166,178 @@ public void getRepresentation_shouldReturnAnInstanceOfNamedRepresentationForGive
...
+	/**
+	 * @verifies fail if failed to get resource classes
+	 * @see RestServiceImpl#getResourceByName(String)
+	 */
+	@Test
+	public void getResourceByName_shouldFailIfFailedToGetResourceClasses() throws Exception {
+		
+		IOException ioException = new IOException("some");
+		
+		when(openmrsClassScanner.getClasses(Resource.class, true)).thenThrow(ioException);
```
dkayiwa: For the above, why not an empty list? That way we are sure that an exception is thrown by the code for a non existing resource.

PR #247, opened 2016-09-27
Convert lineendings from CRLF to LF in xmls

teleivo:
I basically ran

find omod-common/src -exec dos2unix {} \;

in all submodules src folders.

this converted line endings from CRLF to LF in some xml files which have
not been touched since we got rid of the mvn formatter plugin rule
forcing CRLF and use git to handle line endings appropriately.

Now git converts files if necessary from CRLF to LF (see
.gitattributes) this can obfuscate diffs, showing that the entire file
has changed although only line endings changed in every line.
Which is what we've seen in some past commits.

Check with "git show -w" and you'll see that nothing but the line
endings changed in this commit.


dkayiwa:
@teleivo this is causing problems for my omod-1.9/src/test/resources/customTestDataset1_9.xml
Whenever i try to pull, even after a hard reset, i get this:
error: Your local changes to the following files would be overwritten by merge:
	omod-1.9/src/test/resources/customTestDataset1_9.xml
Please, commit your changes or stash them before you can merge.
Aborting

teleivo:
@dkayiwa did you try "git diff -w" when this happens?
it will tell you thats because git changes the line endings from CRLF to LF

dkayiwa:
This is what happens with git diff -w:
warning: CRLF will be replaced by LF in omod-1.9/src/test/resources/customTestDataset1_9.xml.
The file will have its original line endings in your working directory.

teleivo:
@dkayiwa that shows that your customTestDataset1_9.xml has CRLF line endings but they are now stored as LF in ./git objects

```
git rm .gitattributes
git add -A .
git reset --hard HEAD
```

resolves this.

this will happen for all old branches which still have the CRLF line endings.
all new branches will not have this issue.

wluyima:
I would imagine the line endings in our git repo should be already a LF

teleivo:
@wluyima I guess it depends.

If you clone the repo now, git will checkout the files with your local (OS specific) line ending (CRLF, LF). And if you commit any changes it will make sure all files have line endings of LF.

But if you still have old branches on your machine and switch between master and your old branches it will probably tell you that there have been changes and that it will convert CRLF to LF on commit. This is what @dkayiwa was seeing.

PR #249, opened 2016-09-28
Add tests to RestServiceImpl.getSearchHandlers

teleivo:
and getAllSearchHandlers

* add tests to RestServiceImpl.getSearchHandlers
* add tests to RestServiceImpl.getAllSearchHandlers
* adjust test of RestServiceImpl.getSearchHandler to match style used in
all other tests
* fix javadocs for getAllSearchHandlers in RestServiceImpl and RestService

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
@@ -632,16 +632,23 @@ public class RestServiceImpl implements RestService {
 		return resourceHandlers;
 	}
 	
-	/* (non-Javadoc)
-	 * @see org.openmrs.module.webservices.rest.web.api.RestService#getSearchHandlers(java.lang.String)
+	/**
+	 * @see org.openmrs.module.webservices.rest.web.api.RestService#getAllSearchHandlers()
+	 * @should return all search handlers if search handlers have been initialized
+	 * @should return null if search handlers have not been initialized
 	 */
 	public List<SearchHandler> getAllSearchHandlers() {
 		
 		return allSearchHandlers;
 	}
 	
-	/* (non-Javadoc)
+	/**
 	 * @see org.openmrs.module.webservices.rest.web.api.RestService#getSearchHandlers(java.lang.String)
+	 * @should return search handlers for given resource name
+	 * @should return null if no search handler is found for given resource name
+	 * @should return null if no search handler is found for current openmrs version
+	 * @should return null given null
+	 * @should fail if two search handlers for the same resource have the same id
 	 */
 	@Override
 	public Set<SearchHandler> getSearchHandlers(String resourceName) {
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/RestService.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/RestService.java
@@ -81,6 +81,10 @@ public interface RestService {
 	 */
 	public void initialize();
 	
+	/**
+	 * Returns all search handlers.
+	 * 
+	 * @return all search handlers or <code>null</code> if none registered
+	 */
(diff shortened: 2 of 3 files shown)


wluyima:
@teleivo what is the associated ticket number?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -632,16 +632,23 @@ private void eliminateCandidateSearchHandlersWithMissingRequiredParameters(Set<S
 		return resourceHandlers;
 	}
 	
-	/* (non-Javadoc)
-	 * @see org.openmrs.module.webservices.rest.web.api.RestService#getSearchHandlers(java.lang.String)
+	/**
+	 * @see org.openmrs.module.webservices.rest.web.api.RestService#getAllSearchHandlers()
+	 * @should return all search handlers if search handlers have been initialized
```
wluyima: These should be on the service and not the impl

teleivo: what if they are specific to the implementation?
the shoulds describe the behavior of this specific impl.

teleivo:
@wyclif no ticket, just adding tests so the class is covered and I can work on refactoring it and adding new features.

wluyima:
It would be nice to have a ticket, thought there was a ticket related to these methods, you could link to that.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
wluyima: The contract of a service is defined on the interface regardless of the underlying implementation, meaning if we change the implementing class the behavior would still be expected to be the same.

teleivo: Sure, the interface service method signature definitely defines the contract. But it is debatable if some/all/none/which of the @shoulds are part of the service/API contract. Are you only talking about the @shoulds here or in general?

Because for ex. you might not want to force a different impl of the service to always return the Resource with the lowest order if multiple Resources are found for the same class. This allows different impls for the same service.

If you only mean the schoulds here I can move them over to the service interface.

wluyima: Its true that some logic is really specific to an implementation and this is why we have a few tests like XXServiceImplTest or even there are situations where the logic really is DAO specific hence we end up with XXDAOTest classes. But in this case I would think that behavior is part of the service contract.

wluyima: And based on the way @should work if they are on a Service the tests should be in ServiceTest, if you really think this behavior is Impl specific then you would need to move the tests to a new RestServiceImplTest class

teleivo: The annotations are in RestServiceImpl and the tests actually are in RestServiceImplTest

wluyima: Then I guess that's fine

wluyima: Sorry about that :)

wluyima:
Thanks @teleivo ! Next time please create a ticket for the PR :)

dkayiwa:
I think he just did not want to create too many tickets for the lots of refactoring that he is doing. I guess just one ticket should be sufficient for all the refactoring work he is doing.

wluyima:
I agree, i guess the key thing is to link a commit/PR to some ticket which doesn't necessarily have to be new

PR #252, opened 2016-09-29
RESTWS-612 Merge inner wrapper composite key classes into one

teleivo:
two nested classes with almost identical implementation where used as a wrapper
class so two strings could be used as keys to the maps holding SearchHandlers.
first string holds the supported resource of the SearchHandler and second is
either the id or a parameter of the SearchHandler.

* renamed one of them to CompositeSearchHandlerKey
* first property stays supportedResource, second property is simply
secondKey which is once populated with the SearchHandler id and
once with the parameter names.
* use instanceof in equals to check for null and if obj is of the same class

see https://issues.openmrs.org/browse/RESTWS-612


teleivo:
@dkayiwa dont know why the total coverage went down, even if the coverage of the only file that changed went up. I reduced the number of lines, so coverage has to go up.

so if you like the change please merge ;)

will have to find out why this happens to coverage

PR #255, opened 2016-10-26
RESTWS-627 Improve module resource

PawelGutkowski:
https://issues.openmrs.org/browse/RESTWS-627


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
```diff
@@ -52,6 +63,11 @@ public Module newDelegate() {
 	}
 	
 	@Override
+	public Module save(Module delegate) {
+		return delegate;
```
dkayiwa: Do we really support this?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.Collection;
+import java.util.List;
+
+@Resource(name = RestConstants.VERSION_1 + "/moduleaction", supportedClass = ModuleAction.class, supportedOpenmrsVersions = {
+        "1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*" })
+public class ModuleActionResource1_8 extends BaseDelegatingResource<ModuleAction> implements Creatable {
+	
+	private ModuleFactoryWrapper moduleFactoryWrapper = new ModuleFactoryWrapper();
+	
+	public void setModuleFactoryWrapper(ModuleFactoryWrapper moduleFactoryWrapper) {
```
rkorytkowski: Could you please say in comment it is for test purposes?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+	@Override
+	public Object create(SimpleObject post, RequestContext context) throws ResponseException {
+		moduleFactoryWrapper.checkPrivilege();
+		ModuleAction action = parseAction(post, context);
+		Collection<Module> modules = action.isAllModules() ? moduleFactoryWrapper.getLoadedModules() : action.getModules();
+		
+		ServletContext servletContext = getServletContext(context);
+		
+		if (modules.isEmpty()) {
+			throw new IllegalArgumentException("Cannot execute action " + action.getType() + " on empty set of modules.");
```
rkorytkowski: Please throw REST specific exceptions extending ResponseException. We have IllegalPropertyException, but in this case I would create a new one IllegalRequestException.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+	}
+	
+	private ModuleAction parseAction(SimpleObject post, RequestContext context) {
+		
+		//parse action
+		String actionTypeName = post.get("action");
+		ModuleAction.Type actionType;
+		try {
+			if (actionTypeName == null) {
+				throw new IllegalArgumentException("Please specify action type");
```
rkorytkowski: Throw IllegalRequestException

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+			}
+			actionType = ModuleAction.Type.valueOf(actionTypeName.toUpperCase().trim());
+		}
+		catch (Exception e) {
+			throw new IllegalArgumentException("Invalid modules action", e);
```
rkorytkowski: Throw IllegalRequestException. No need to rethrow "Please specify action type", make the try block smaller.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+		
+		//instanceof returns false if null
+		if (allModulesValue instanceof Boolean) {
+			allModules = (Boolean) allModulesValue;
+		} else if (allModulesValue instanceof String) {
+			allModules = Boolean.valueOf((String) allModulesValue);
+		}
+		
+		if (allModules && !modules.isEmpty()) {
+			throw new IllegalArgumentException("Please choose either 'allModules' or modules");
```
rkorytkowski: Throw IllegalRequestException.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+	}
+	
+	/**
+	 * Overriding create directly, because ModuleFactory requires ServletContext to execute any
+	 * action
+	 */
+	@Override
+	public Object create(SimpleObject post, RequestContext context) throws ResponseException {
+		moduleFactoryWrapper.checkPrivilege();
+		ModuleAction action = parseAction(post, context);
```
rkorytkowski: Why don't you just do what the original create method does i.e. call newDelegate(); and setConvertedProperties(delegate, post, getCreatableProperties(), true); ?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+	}
+	
+	@Override
+	public ModuleAction newDelegate() {
+		return new ModuleAction();
+	}
+	
+	@Override
+	public ModuleAction save(ModuleAction delegate) {
+		throw new IllegalArgumentException("ModuleAction cannot be saved");
```
rkorytkowski: We should actually say here that since create() is overwritten this method must not be called ever... so probably UnsupportedOperationException would be better.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
```diff
@@ -20,25 +20,36 @@
...
 @Resource(name = RestConstants.VERSION_1 + "/module", supportedClass = Module.class, supportedOpenmrsVersions = { "1.8.*",
         "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*" })
-public class ModuleResource1_8 extends BaseDelegatingReadableResource<Module> {
+public class ModuleResource1_8 extends BaseDelegatingReadableResource<Module> implements Uploadable {
+	
+	private ModuleFactoryWrapper moduleFactoryWrapper = new ModuleFactoryWrapper();
+	
+	private String moduleActionLink = ModuleActionResource1_8.class.getAnnotation(Resource.class).name();
+	
+	public void setModuleFactoryWrapper(ModuleFactoryWrapper moduleFactoryWrapper) {
```
rkorytkowski: Please state it's for testing.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
```diff
@@ -103,4 +128,43 @@ public static String getDisplay(Module instance) {
...
+	@Override
+	public Object upload(MultipartFile file, RequestContext context) throws ResponseException, IOException {
+		moduleFactoryWrapper.checkPrivilege();
+		
+		File moduleFile = null;
+		Module module = null;
+		
+		try {
+			if (file == null || file.isEmpty()) {
+				throw new IllegalArgumentException("Uploaded OMOD file cannot be empty");
```
rkorytkowski: Throw IllegalRequestException

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleAction.java
```diff
@@ -0,0 +1,60 @@
...
+ */
+package org.openmrs.module.webservices.helper;
+
+import org.openmrs.module.Module;
+
+import java.util.Collection;
+
+public class ModuleAction {
+	
+	public enum Type {
```
dkayiwa: I would call this enum Action instead of Type. What do you think @PawelGutkowski? :)

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+		
+		if (needsRefresh) {
+			moduleFactoryWrapper.refreshWebApplicationContext(servletContext);
+		}
+	}
+	
+	private void stopModules(Collection<Module> modules, ServletContext servletContext) {
+		for (Module module : modules) {
+			if (moduleFactoryWrapper.isModuleStarted(module)) {
+				moduleFactoryWrapper.stopModule(module, servletContext);
```
dkayiwa: The above refreshes the spring application context for each module. Was this your intention?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+				case UNLOAD:
+					unloadModules(modules, servletContext);
+					break;
+			}
+		}
+		
+		return ConversionUtil.convertToRepresentation(action, Representation.DEFAULT);
+	}
+	
+	private ModuleAction parseAction(SimpleObject post, RequestContext context) {
```
wluyima: Why parse, action should be an enumaredt field on the ModuleAction class

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
```diff
@@ -103,4 +128,43 @@ public static String getDisplay(Module instance) {
 		return instance.getName();
 	}
 	
+	@Override
+	public Object upload(MultipartFile file, RequestContext context) throws ResponseException, IOException {
```
wluyima: How is this invoked by the framework?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+import org.openmrs.module.webservices.rest.web.response.ResponseException;
+
+import javax.servlet.ServletContext;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
+@Resource(name = RestConstants.VERSION_1 + "/moduleaction", supportedClass = ModuleAction.class, supportedOpenmrsVersions = {
+        "1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*" })
+public class ModuleActionResource1_8 extends BaseDelegatingResource<ModuleAction> implements Creatable {
```
dkayiwa: I do not even see Updatable fitting in the picture. Unless the class was named something like ModuleStateResource. Not a big deal though. So i agree that Updatable is more suitable than Creatable. :)

omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/MockModuleFactoryWrapper.java
```diff
@@ -0,0 +1,110 @@
...
+
+import javax.servlet.ServletContext;
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+
+public class MockModuleFactoryWrapper extends ModuleFactoryWrapper {
```
wluyima: Why do you need this wrapper?

wluyima:
I feel like this is way more code that what is actually needed to get this to work, for instance I don't think the wrapper is needed, we should stick closer to the how other resources are written and not try to bake a lot of custom logic in this resource.

You need to add a ModuleAction class with the modules and action fields, add a resource for it and everything will work magically

wluyima:
I implemented something similar to the module action in the case reporting module, you can take a look at the [StatusChange](https://github.com/openmrs/openmrs-module-casereport/blob/master/omod/src/main/java/org/openmrs/module/casereport/web/rest/StatusChange.java) class which is similar to a module action, then I added a [StatusChangeResource](https://github.com/openmrs/openmrs-module-casereport/blob/master/omod/src/main/java/org/openmrs/module/casereport/web/rest/v1_0/resource/StatusChangeResource.java) for it which leaves the code minimal and similar to other resources

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+		if (!errors.isEmpty()) {
+			StringBuilder stringBuilder = new StringBuilder();
+			for (Exception error : errors) {
+				stringBuilder.append(error.getMessage()).append("; ");
+			}
+			throw new RuntimeException(stringBuilder.toString());
+		}
+	}
+	
+	private Module findByUuid(String uuid) {
```
wluyima: Do modules have uuid?

wluyima: You should implement getByUniqueId() just like other resources, it also has the benefit of the appropriate exception getting thrown if no match is found instead of the exception you're throwing below

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
```diff
@@ -103,4 +128,43 @@ public static String getDisplay(Module instance) {
...
+		moduleFactoryWrapper.checkPrivilege();
+		
+		File moduleFile = null;
+		Module module = null;
+		
+		try {
+			if (file == null || file.isEmpty()) {
+				throw new IllegalArgumentException("Uploaded OMOD file cannot be empty");
+			} else {
+				String filename = WebUtil.stripFilename(file.getOriginalFilename());
```
wluyima: I'd imagine you borrowed most of this logic from the ModuleListController in the legacyui module.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,245 @@
...
+	private void restartModules(Collection<Module> modules, ServletContext servletContext) {
+		for (Module module : modules) {
+			if (moduleFactoryWrapper.isModuleStarted(module)) {
+				moduleFactoryWrapper.stopModuleSkipRefresh(module, servletContext);
+			}
+		}
+		
+		boolean needsRefresh = false;
+		for (Module module : moduleFactoryWrapper.getModulesInStartupOrder(modules)) {
+			if (moduleFactoryWrapper.isModuleStopped(module)) {
```
PawelGutkowski: Thanks for heads up, I will look into this!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
PawelGutkowski: No, it was by mistake, I've removed it, thanks!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: Done

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: Yeah, I probably should do it from the start, I forgot I could use those methods here... I've fixed it, now custom ModuleAction deserialization logic is no longer needed. Thanks!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: Ok, I've added `IllegalRequestException` class and replaced `IllegalArgumentException` here and in other places.

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleFactoryWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+	 */
+	public void stopModuleSkipRefresh(Module module, ServletContext servletContext) {
+		try {
+			Method stopModule = WebModuleUtil.class.getDeclaredMethod("stopModule", Module.class, ServletContext.class,
+			    Boolean.TYPE);
+			stopModule.setAccessible(true);
+			stopModule.invoke(null, module, servletContext, true);
+		}
+		catch (NoSuchMethodException e) {
+			throw new RuntimeException();
```
PawelGutkowski: True, I've added messages and caught exception to constructor

omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_8/ModuleActionController1_8Tests.java
```diff
@@ -0,0 +1,162 @@
...
+		        + getUuid() + "\"]}")));
+		assertThat(mockModuleFactory.loadedModules, not(hasItem(atlasModule)));
+	}
+	
+	@Test
+	public void shouldStartAllModules() throws Exception {
+		SimpleObject deserialized = deserialize(handle(newPostRequest(getURI(),
+		    "{\"action\":\"start\", \"allModules\":\"true\"}")));
+		
+		for (Module loadedModule : mockModuleFactory.loadedModules) {
```
PawelGutkowski: With way how @Before is written, it cannot be other way, but sanity check won't hurt, I've added assertion

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleAction.java
```diff
@@ -0,0 +1,60 @@
...
+	
+	public ModuleAction(Type action, Collection<Module> modules, boolean startAll) {
+		this.modules = modules;
+		this.type = action;
+		this.allModules = startAll;
+	}
+	
+	private Collection<Module> modules;
+	
+	private Type type;
```
PawelGutkowski: Okay, done

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
PawelGutkowski: I've added comment as requested

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleAction.java
PawelGutkowski: Fine by me, done :)

omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/MockModuleFactoryWrapper.java
PawelGutkowski: For testing purposes: all operations on modules are executed via `ModuleFactory` static methods, `Module` class is final, and actions like WebContextRefresh are failing in testing environment. Creating wrapper for `ModuleFactory` and mocking it was the easiest way to test ModuleActionResource.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: After changes suggested by @rkorytkowski this method is no longer in PR, as now deserialization is handled by methods from `BaseDelegatingResource` superclass

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
PawelGutkowski: Yes

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
```diff
@@ -252,7 +252,7 @@ public static Object convert(Object object, Type toType) throws ConversionExcept
 			} else if (toClass.isAssignableFrom(Locale.class)) {
 				return LocaleUtility.fromSpecification(object.toString());
 			} else if (toClass.isEnum()) {
-				return Enum.valueOf((Class<? extends Enum>) toClass, object.toString());
+				return Enum.valueOf((Class<? extends Enum>) toClass, object.toString().toUpperCase());
```
PawelGutkowski: @rkorytkowski @dkayiwa what do you think about it? I think it is more convenient to convert enums case-insensitively

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: @dkayiwa I found out that there was one method invocation missing in `ModuleFactoryWrapper`, thanks!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+		if (refreshContext) {
+			moduleFactoryWrapper.refreshWebApplicationContext(servletContext);
+		}
+	}
+	
+	/**
+	 * @param modules modules to start
+	 * @param servletContext ServletContext is required by WebModuleUtil to perform operation
+	 */
+	private void startModules(Collection<Module> modules, ServletContext servletContext) {
```
PawelGutkowski: I found out, it fails whenever metadatamapping module is started with other modules. metadatamapping module requires REST module and have components which extend classes from REST module, I think it has something to do with it.

dkayiwa: @PawelGutkowski so did you eventually fix this?

PawelGutkowski: @dkayiwa Unfortunately I didn't yet.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+	}
+	
+	@Override
+	public ModuleAction newDelegate() {
+		return new ModuleAction();
+	}
+	
+	@Override
+	public ModuleAction save(ModuleAction delegate) {
+		throw new UnsupportedOperationException("ModuleAction cannot be saved");
```
wluyima: This should be ResourceDoesNotSupportOperationException, if the swagger creator checks another class that was a mistake and we need to fix the swagger creator instead to check the correct exception

omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/MockModuleFactoryWrapper.java
wluyima: My fault! I didn't realize this was in the test scope, that's fine I guess

wluyima:
I see that uuids in ModuleResource are generated using the name and not the moduleId, anybody knows why it's the case?

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleFactoryWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Collection;
+import java.util.List;
+
+/**
+ * ModuleFactoryWrapper wraps static methods of ModuleFactory to make code operating on modules
+ * testable
+ */
+public class ModuleFactoryWrapper {
```
wluyima: I'm okay we having this wrapper for methods that delegate multiple calls to other methods in core but I don't think you need methods on this wrapper that make a single call directly to the ModuleFactory or WebModuleUtil or ModuleUtil.

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleFactoryWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+	public void refreshWebApplicationContext(ServletContext context) {
+		WebModuleUtil.refreshWAC(context, false, null);
+	}
+	
+	public Collection<Module> getLoadedModules() {
+		return ModuleFactory.getLoadedModules();
+	}
+	
+	public boolean isModuleStarted(Module module) {
+		return module.isStarted();
```
wluyima: like this is not a needed method because it really has logic of is own

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleFactoryWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+	public void unloadModule(Module module) {
+		ModuleFactory.unloadModule(module);
+	}
+	
+	public File insertModuleFile(Module module, String filename) throws FileNotFoundException {
+		return ModuleUtil.insertModuleFile(new FileInputStream(module.getFile()), filename);
+	}
+	
+	public Module loadModule(File moduleFile) {
+		return ModuleFactory.loadModule(moduleFile);
```
wluyima: Same for this method

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
wluyima: But I see several variations in the logic

wluyima: E.g I don't see where in the legacyui they explicitly deal with dependent module like you are doing here, isn't that taken care of by the API?

wluyima: Or it's a new feature we need to support via REST?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+			modules = action.getModules();
+		}
+		
+		ServletContext servletContext = getServletContext(context);
+		
+		if (modules == null || modules.isEmpty()) {
+			throw new IllegalRequestException("Cannot execute action " + action.getAction() + " on empty set of modules.");
+		} else {
+			//rest module cannot be stopped/unloaded during handling request
+			Module restModule = moduleFactoryWrapper.getModuleById(RestConstants.MODULE_ID);
```
wluyima: Nice anticipation!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+		stopModules(modules, servletContext, false);
+		startModules(modules, servletContext);
+	}
+	
+	private void unloadModules(Collection<Module> modules, ServletContext servletContext) {
+		boolean needsRefresh = false;
+		for (Module module : modules) {
+			if (moduleFactoryWrapper.isModuleStarted(module)) {
+				moduleFactoryWrapper.stopModuleSkipRefresh(module, servletContext);
+				needsRefresh = true;
```
wluyima: Do you really need a refresh when a module is unloaded? In ModuleListController in leacyui it doesn't seem to be done. I wonder if it was an oversight in the legacyui module or if it's actually unnecessary

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+	 * @param servletContext ServletContext is required by WebModuleUtil to perform operation
+	 */
+	private void stopModules(Collection<Module> modules, ServletContext servletContext, boolean refreshContext) {
+		for (Module module : modules) {
+			if (moduleFactoryWrapper.isModuleStarted(module)) {
+				moduleFactoryWrapper.stopModuleSkipRefresh(module, servletContext);
+			}
+		}
+		if (refreshContext) {
+			moduleFactoryWrapper.refreshWebApplicationContext(servletContext);
```
wluyima: Same here, is the refresh necessary?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+				}
+			}
+		}
+		
+		if (!errors.isEmpty()) {
+			StringBuilder stringBuilder = new StringBuilder();
+			for (Exception error : errors) {
+				stringBuilder.append(error.getMessage()).append("; ");
+			}
+			throw new RuntimeException(stringBuilder.toString());
```
wluyima: Does this have to fail? Or just report the error message? I think in the legacy ui module we don't do this

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
wluyima: Personally, I would move code from the create() method above to this method and get rid of create, that way you don't have to deal with both

wluyima: Though one might argue that this action is more of a create than a save, may be you can keep it this way, I'm fine with it too, I guess its up to you.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
```diff
@@ -20,25 +20,36 @@
...
+	
+	public void setModuleFactoryWrapper(ModuleFactoryWrapper moduleFactoryWrapper) {
+		this.moduleFactoryWrapper = moduleFactoryWrapper;
+	}
 	
 	@Override
 	public Module getByUniqueId(String uniqueId) {
-		for (Module module : ModuleFactory.getLoadedModules()) {
+		for (Module module : moduleFactoryWrapper.getLoadedModules()) {
 			if (uniqueId.equals(getUuid(module))) {
```
wluyima: I see that the ModuleResource1_8.getUuid() was added to generate a uuid on the fly using the module name, that was a mistake and may be we should fix it. UUID doesn't have to be a uuid in its literal sense, in our REST API uuid can be any unique identifier which in this case should be the module id in OpenMRS core. I think we need to use the value returned by module.getModuleId() instead of generating some arbitrary uuids.

@bmamlin  can weigh in on this

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
dkayiwa: @PawelGutkowski can you confirm that moduleFactoryWrapper.isModuleStopped () returns true for each of the modules?
By the way, how are you reproducing this error? From a unit test?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
PawelGutkowski: legacyui does it, see if statement body from https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/main/java/org/openmrs/module/web/controller/ModuleListController.java#L128

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: I'm overwriting `create` method because to execute any operation on module, ServletContext has to be passed, it is commented in code: https://github.com/PawelGutkowski/openmrs-module-webservices.rest/blob/460018808b7bbff61d3405ea916346b1370b84c8/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java#L49

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleFactoryWrapper.java
PawelGutkowski: Single calls need to be wrapped too, because they fail in test context, and that's why `MockModuleFactoryWrapper` overwrites them

omod-common/src/main/java/org/openmrs/module/webservices/helper/ModuleFactoryWrapper.java
PawelGutkowski: see https://github.com/openmrs/openmrs-module-webservices.rest/pull/255#discussion_r85474581

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+	 */
+	@Override
+	public Object create(SimpleObject post, RequestContext context) throws ResponseException {
+		moduleFactoryWrapper.checkPrivilege();
+		
+		ModuleAction action = newDelegate();
+		setConvertedProperties(action, post, getCreatableProperties(), true);
+		
+		Collection<Module> modules;
+		if (action.isAllModules() != null && action.isAllModules()) {
```
PawelGutkowski: no, it is instance of  `Boolean` and it would throw `NullPointerException` if it was just `(action.isAllModules)` without null check.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
```diff
@@ -0,0 +1,225 @@
...
+				case RESTART:
+					restartModules(modules, servletContext);
+					break;
+				case UNLOAD:
+					unloadModules(modules, servletContext);
+					break;
+			}
+		}
+		
+		return ConversionUtil.convertToRepresentation(action, Representation.DEFAULT);
```
PawelGutkowski: Client is just sending a couple of uuids and action string, i thought it would be nice to return a bit more info about affected modules, similiary how we return created object in response body when handling POST request.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: I think it is important to report to client that requested action didn't succeed

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
PawelGutkowski: I just decided keep convention which was already in place in `ModuleResource1_8`. If we decide it needs to be changed, fine by me.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
dkayiwa: @PawelGutkowski are you able to reproduce this from the web interface when you click the Start All button on the manage modules screen?
If not able to reproduce, can you look at the controller to see if it implements start all in the same way as you are doing?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
wluyima: The convention really has nothing to do with the uuid but has a lot to do with a resource having a unique identifier which is the moduleId for a module

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
wluyima: I thought under the hood you delegate calls to Modulefactory.startModule and stopModule meaning they will anyways take care of the refresh if needed

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
wluyima: action is constructed from what the user submitted meaning that you are sending back the same data, is that useful?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
wluyima: to fix that don't you just need to convert the type to `boolean` from `Boolean` that way you don't need the null check here

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
wluyima: Ok

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: @dkayiwa starting all modules old interface works without problem. It seems to me that the procedure is the same - order modules in startup order, start with ModuleFactory, start with WebModuleUtil (without WAC refresh), refresh WAC.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
PawelGutkowski: yes, but when starting multiple modules, it makes no sense to refresh after every module start

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
dkayiwa: @PawelGutkowski i have just installed the modules you mentioned above, stopped metadatasharing, and then stopped metadatamapping. Then i sent a post request with the body below: 
{   "allModules": "true",   "action": "start" }
It has worked seamlessly. Is there anything am missing?

PawelGutkowski: No, you reproduced my environment. I tried it many times on 3 different servers, but maybe something is wrong with my setup... thanks for checking it out!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
wluyima: I see, in the legacy ui we didnt have this start multiple feature, so that makes sense then

rkorytkowski: @wluyima, we did have it, see https://issues.openmrs.org/browse/TRUNK-4438

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleActionResource1_8.java
rkorytkowski: Explained by @PawelGutkowski above.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
rkorytkowski: @wluyima, could you please bring it up on talk? :) It's an important design decision, which shouldn't be made on a pull request.

rkorytkowski:
Thanks @wluyima and @dkayiwa for reviewing! Most of all thanks @PawelGutkowski for working on this! We need the code in another project so I'll merge it in (see https://github.com/openmrs/openmrs-module-openconceptlab/pull/37). The only issue remaining I noticed is using uuids generated from moduleId, which @wluyima thinks is wrong. We'll discuss that on talk and address as a follow up if needed.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ModuleResource1_8.java
wluyima: Done! See https://talk.openmrs.org/t/module-identifiers-in-restws-module/8641

PR #261, opened 2016-12-06
RESTWS-613: Provide SearchHandelr for subclasses 

lkukielka:


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -97,54 +98,27 @@ public ResourceDefinition(Resource resource, int order) {
...
-				if (other.secondKey != null)
-					return false;
-			} else if (!secondKey.equals(other.secondKey))
-				return false;
-			if (supportedResource == null) {
-				if (other.supportedResource != null)
-					return false;
-			} else if (!supportedResource.equals(other.supportedResource))
-				return false;
-			return true;
```
rkorytkowski: hashCode and equals are essential here and shouldn't be removed...

dkayiwa:
@lkukielka did you see the travis failure?

lkukielka:
@dkayiwa Yes, I did. It is not final solution. I'm still working on this.

lkukielka:
@rkorytkowski Could you take a look at this? I can't localize error, probably in  RestServiceImpl.java, which causes a test failure.

rkorytkowski:
@lkukielka, why don't you paste stacktrace or something relevant to the error so that I don't have to look through the whole build log?

Should I look at the travis failure? Is it the error you mean:

Failed tests:   shouldFindUserByUsername(org.openmrs.module.webservices.rest.web.v1_0.controller.openmrs1_8.UserController1_8Test): expected:<1> but was:<3>

It finds more users than expected. What are the users and what are the results?

lkukielka:
@rkorytkowski 
Yes, I meant this error. As u said it finds more users than expected. Actually it finds all users, but it should find one specific user. 

I think the current solution is not working as I wish. I noticed that the reason for this bug is that I cant get proper searchHandler in the following line https://github.com/lkukielka/openmrs-module-webservices.rest/blob/a3b6565c45e001c93f104096e7397faaa147033c/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java#L549 becouse there are search handlers without a value but I'm looking for search handler with value, so searchHandlersByParameter.get( ... ) returns null

omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_8/UserController1_8Test.java
```diff
@@ -245,6 +291,17 @@ public void shouldFindUserByUsername() throws Exception {
 	}
 	
 	@Test
+	public void shouldFindUserByUsernameUsingSearchParameter() throws Exception {
+		SimpleObject response = deserialize(handle(newGetRequest(getURI(), new Parameter("username", "admin"))));
```
rkorytkowski: You do not test here that the correct handler was picked up to do the search. I would change the test to for example:
shouldNotFindAdminByUsername and make the mocked handler return empty list when admin is requested.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/SubresourceParameter.java
```diff
@@ -0,0 +1,18 @@
...
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.resource.api;
+
+public class SubresourceParameter extends SearchParameter {
+	
+	SubresourceParameter(String value) {
```
rkorytkowski: Make the constructor public

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/SearchParameter.java
```diff
@@ -0,0 +1,65 @@
...
+	
+	public void setValue(String value) {
+		this.value = value;
+	}
+	
+	@Override
+	public boolean equals(Object o) {
+		if (this == o)
+			return true;
+		if (o == null || getClass() != o.getClass())
```
rkorytkowski: What if you use SubresourceParameter? You should use instanceof here.

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImplTest.java
```diff
@@ -637,12 +638,13 @@ public void getSearchHandler_shouldReturnSearchHandlerProvidingAllRequestParamet
...
 		SearchConfig searchConfig1 = new SearchConfig("default", "v1/concept", "1.8.*", new SearchQuery.Builder(
-		        "description").withRequiredParameters("source").withOptionalParameters("code").build());
+		        "description").withRequiredParameters(new SearchParameter("source", "some source"))
+		        .withOptionalParameters(new SearchParameter("code", "some code")).build());
 		when(searchHandler1.getSearchConfig()).thenReturn(searchConfig1);
 		
 		SearchHandler searchHandler2 = mock(SearchHandler.class);
 		SearchConfig searchConfig2 = new SearchConfig("conceptByMapping", "v1/concept", "1.8.*", new SearchQuery.Builder(
-		        "description").withRequiredParameters("source").build());
+		        "description").withRequiredParameters(new SearchParameter("source", "some source")).build());
```
lkukielka: Yes. This test was failing, but it depends on changes made in SearchQuery, exactly on type of required and optionalParameters, so there I needed to change test.

rkorytkowski: You should still be able to call withRequiredParameters("source")... I don't understand.

lkukielka: Okay, I checked it and you are right. It works now. I'm going to revert these tests.

lkukielka:
@rkorytkowski I fixed above things and also added test for _optionalRequestParameter._

omod-1.8/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_8/UserController1_8Test.java
```diff
@@ -245,6 +313,28 @@ public void shouldFindUserByUsername() throws Exception {
...
+	@Test
+	public void shouldFindUserByUsernameUsingOptionalSearchParameterForMockedHandler() throws Exception {
+		SimpleObject response = deserialize(handle(newGetRequest(getURI(), new Parameter("username", "bruno"))));
+		List<Object> results = Util.getResultsList(response);
+		
+		Object next = results.iterator().next();
+		
+		assertEquals(1, results.size());
+		assertEquals(MOCKED_USER_NAME_FOR_OPTIONAL_PARAM, PropertyUtils.getProperty(next, "display"));
+	}
```
rkorytkowski: Some ideas for tests:
Define handlers:
1) required username:admin, optional preferredLocales
2) required systemId, optional username:bruno
3) optional username:bruno

1. Match 1) for query "?username=admin&preferredLocales=en"
2. Match 2) for query "?systemId=bruno"
3. Match 2) for query "?systemId=bruno&username=bruno"
4. Match 3) for query "?username=bruno"
5. Do not match 1) for "?username=bruno&preferredLocales=en"
6. Do not match 3) for "?username=james"
7. Do not match 2) for "?username=james&systemId=bruno"

PR #276, opened 2017-03-29
RESTWS-653 Bad use of operations on String

gayanW:


omod-common/src/main/java/org/openmrs/module/webservices/docs/SearchHandlerDoc.java
```diff
@@ -94,13 +95,8 @@ public void setSearchQueriesDoc(List<SearchQueryDoc> searchQueriesDoc) {
 	}
 	
 	private String extractResourceName(String supportedResourceUrl) {
-
```
dkayiwa: How about just removing this method? For it is not used anywhere. Not so?

gayanW: Yes, It is not used within the module. I could either remove the method or close the PR as it is not that important. What do you think?

dkayiwa: Just remove the method.

PR #278, opened 2017-04-12
RESTWS-561 Document Basic Search

gayanW:
Adds GET operation to root path of a resource, when the resource does not have
a getAll or a custom search handler, but has got doSearch implemented


dkayiwa:
@psbrandt what do you think of this?

dkayiwa:
@psbrandt do you have any objections to merging this? 😊

PR #282, opened 2017-06-06
RESTWS-662 add methods: getResName(), getResParentName()

gayanW:


omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -910,7 +894,65 @@ private void addPaths() {
...
+		return resourceName;
+	}
+
+	/**
+	 * Extracts resource parent name from a resource handler
+	 * @param trim if false returns name with version, if true version part of the name is trimmed
+	 * @return name of the parent resource or null if its not a sub-resource
+	 */
+	private String getResourceParentName(DelegatingResourceHandler resourceHandler, boolean trim) {
+		SubResource subResourceAnnotation = resourceHandler.getClass().getAnnotation(SubResource.class);
```
wluyima: There should be a way to minimize this duplication by making this method delegate to the one above

gayanW: I can't seem to figure out easy way of delegating the two. Though two looks similar, logic there is quite different. GetResourceParentName() only do a one conditioning where as getResourceName() check for a Resource, and also for a SubResource.

I could however combine them into a single method, but it seems to degrade the readability. Please see: https://gist.github.com/gayanW/2df142a6bddffd59b04b8982c10d1985

wluyima: The only difference between the 2 is that one also checks if it's a sub resource, it means you can have one method that takes in an extra boolean argument that specifies if you are getting a parent's name or not, this is the same argument in the single method you should use to determine whether you should assume it's a sub resource.

gayanW: Sorry I'm still confused. If we are to add a boolean like `getParent` then we have to assume that the caller knows if its SubResource or not before calling the method. I can't seem to get things to delegate without making things extra complicated. 

I wouldn't mind you making a commit or gist, if you've got the time of course! Thanks.

wluyima: Think about it this way, by the time one calls `getResourceParentName()`, they already know that the resource handler is a sub resource, otherwise how do they determine which one of your 2 methods to call in your current code? In fact, I kind of agree with you that client code should not have to know if they are dealing with a resource vs sub resource, this makes me think that you should actually have one method and in your code you should check if resource handler is a resource vs a sub resource via reflection

gayanW: So what about my first solution. It's some what efficient as well. getResName() can be used for both Res and SubRes. getResName() uses reflection to check whether it a Resource or a SubResource where as getResourceParentName() pre-assumes that it is a SubResource. If it's not a SubResource then it returns null.

That would be great that if we could get one to delegate other, but getting name and the parent have different conditioning order. So would you want me to keep my current commit or to make a new commit following the single method approach as shown in gist?

dkayiwa: @wluyima any response to the above?

wluyima: What I said was just a suggestion for cleaner code but I know the code works as it is

gayanW: @dkayiwa @wluyima i created this as a helper to  RESTWS-56 which I'm working on now, May we keep this let alone for a while..? till I actually tested this with RESTWS-56, i'll ping you guys once ready, so we could make changes as necessary and merge it.

dkayiwa: @gayanW did you just forget to ping us? Or you are not yet ready? 😊

dkayiwa: @gayanW did you see this?

gayanW: @dkayiwa Please would you wait a bit longer until RESTWS-562 is being completed :blush:

dkayiwa: Sure no problem. I was just checking if you still have it in mind. 😊

dkayiwa: @gayanW now that RESTWS-562 is merged, is it time to revisit this?

gayanW: With the final changes made to RESTWS-562, these methods will not be reused. So I think, we better not add more code. @dkayiwa @wluyima Sorry for all the time that it took to review. :) 

May be we should close the ticket or move it to backlog?

dkayiwa: Feel free to close it.

PR #283, opened 2017-06-20
RESTWS-395: UUID preferred over name when setting PIT.

mks-d:


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java
```diff
@@ -89,11 +90,12 @@ public DelegatingResourceDescription getCreatableProperties() {
 	public void setIdentifierType(PatientIdentifier instance, PatientIdentifierType identifierType) {
 		String uuid = identifierType.getUuid();
 		String name = identifierType.getName();
-		if (name != null && !name.isEmpty()) {
-			instance.setIdentifierType(Context.getPatientService().getPatientIdentifierTypeByName(name));
-		} else if (uuid != null && !uuid.isEmpty()) {
+		if (!StringUtils.isEmpty(uuid)) {
 			instance.setIdentifierType(Context.getPatientService().getPatientIdentifierTypeByUuid(uuid));
 		}
+		if (instance.getIdentifierType() == null && !StringUtils.isEmpty(name)) {
```
dkayiwa: The original code never had a check for getIdentifierType() being null. Could this be out of scope for the commit message and intention?

mks-d: It's slightly out of scope, but basically it is about trying the fetching by name only if the fetching by UUID had failed.
The above code basically tries everything to the best of the available possibilities to set the PIT.

dkayiwa: The original code set the passed in identifier type even when instance.getIdentifierType() was not null. Isn't this changing that behaviour? I just do not want us introduce any unintentional bugs. I thought all you wanted for this pr was just reverse the order.

wluyima: I think the change looks good to me, all he's doing is to find the identifier type by uuid or name and that's why the check for null becomes necessary

wluyima: The only strange thing am seeing is, why do we have to fetch the identifier type object from the service again when we already have an instance anyways?

mks-d: @dkayiwa, can this be merged then?

dkayiwa: If uuid is null, name is not null, and instance.getIdentifierType() is not null, but has a value that is not represented by the identifier type in the supplied name, would this code do the expected?

mks-d: I believe so since using the setter is meant to override the PIT member of `instance`.
Yes that would be the expected behavior.

dkayiwa: The original code would call instance.setIdentifierType() with the identifier type as represented by name. With your changes, as long as instance.getIdentifierType() is not null, instance.setIdentifierType() would not be called to set the identifier type in name.

mks-d: Ah sure.
Anyway I'll close this since I need to switch branches on our fork. We'll follow up on the new PR.
Thanks for your analysis.

PR #294, opened 2017-08-26
RESTWS-568 - Optimize Swagger spec creation

gayanW:
Generate and cache Swagger spec and clear it once every context refresh. If a cache exists SwaggerSpecificationCreator will serve the cached version instead of regenerating the spec.

Issue: https://issues.openmrs.org/browse/RESTWS-568

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
@@ -73,7 +73,7 @@ import java.util.Set;
 
 public class SwaggerSpecificationCreator {
 	
-	private Swagger swagger = new Swagger();
+	private static Swagger swagger;
 	
 	private String host;
 	
@@ -116,7 +116,10 @@ public class SwaggerSpecificationCreator {
 		return this;
 	}
 	
-	public String BuildJSON() {
+	/**
+	 * Regenerate the swagger spec from scratch
+	 */
+	private void BuildJSON() {
 		synchronized (this) {
 			log.info("Initiating Swagger specification creation");
 			toggleLogs(RestConstants.SWAGGER_LOGS_OFF);
@@ -133,6 +136,16 @@ public class SwaggerSpecificationCreator {
 				toggleLogs(RestConstants.SWAGGER_LOGS_ON);
 				log.info("Swagger specification creation complete");
 			}
+		}
+	}
+	
+	public String getJSON() {
+		if (isCached()) {
+			log.info("Returning a cached copy of Swagger specification");
+			initSwagger();
+		} else {
+			swagger = new Swagger();
+			BuildJSON();
 		}
 		return createJSON();
 	}
@@ -200,7 +213,7 @@ public class SwaggerSpecificationCreator {
 		        .contact(new Contact().name("OpenMRS").url("http://openmrs.org"))
 		        .license(new License().name("MPL-2.0 w/ HD").url("http://openmrs.org/license"));
 		
-		this.swagger
+		swagger
 		        .info(info)
 		        .host(this.host)
 		        .basePath(this.basePath)
@@ -849,12 +862,10 @@ public class SwaggerSpecificationCreator {
 					parameter.setRequired(false);
 					parameters.add(parameter);
 				}
-				
 				break;
 			}
 		}
 		return parameters;
-		
 	}
 	
 	private String createJSON() {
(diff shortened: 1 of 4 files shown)


omod/src/main/java/org/openmrs/module/webservices/rest/Activator.java
```diff
@@ -28,6 +29,9 @@
 	@Override
 	public void started() {
 		log.info("Started the REST Web Service module");
+		
+		// Cache Swagger Spec
+		new SwaggerSpecificationCreator("host/openmrs/ws/rest").BuildJSON();
```
gayanW: If we could get the `host` and the `context path` in here, we could simplify this PR a little more.

dkayiwa: Did you intentionally hard code the above?

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -93,26 +93,38 @@ public SwaggerSpecificationCreator(String baseUrl) {
...
-				log.error("Error while creating Swagger specification", e);
-			}
-			finally {
-				toggleLogs(RestConstants.SWAGGER_LOGS_ON);
-				log.info("Swagger specification creation complete");
+		if (swagger != null) {
+			updateHostAndBasePath();
+		} else {
+			synchronized (this) {
+				log.info("Initiating Swagger specification creation");
```
dkayiwa: Wrapping the else body in a new method will make it even cleaner.

omod/src/main/java/org/openmrs/module/webservices/rest/Activator.java
gayanW: Yes. I don't think we could get the actual host and the context path there. Let me know if we could. So therefore we cache it with some hard coded string as the base uri. Thereafter whenever some one call new `SwaggerSpecificationCreator("updated_string").BuildJSON()` (as done in [here](https://github.com/gayanW/openmrs-module-webservices.rest/blob/a72ba43ecf7392dd22747671a5071f47d02f56ae/omod/src/main/java/org/openmrs/module/webservices/rest/web/controller/SwaggerSpecificationController.java#L31-L65)) it will return the cached but with the `updated_string`.

dkayiwa: Did you try "WebConstants.WEBAPP_NAME" to get the context path?

dkayiwa:
@gayanW do you see the merge conflicts here?

omod/src/main/java/org/openmrs/module/webservices/rest/Activator.java
gayanW: We need the name of the running host (domain).

gayanW:
Conflicts were fixed.

omod/src/main/java/org/openmrs/module/webservices/rest/Activator.java
dkayiwa: Since this is running from the server itself, can't you just use localhost?

gayanW: No then the client (try out) request will be sent to his own machines localhost. May we wait until we come to a conclusion on https://talk.openmrs.org/t/cant-make-requests-from-the-api-documentation/13145/11?u=gayanw

Then we could omit the `host/..` string altogether.
@psbrandt What about removing the host param from the spec?

dkayiwa:
@gayanW are you aware of the merge conflicts on this?

omod/src/main/java/org/openmrs/module/webservices/rest/Activator.java
```diff
@@ -28,6 +29,9 @@
 	@Override
 	public void started() {
 		log.info("Started the REST Web Service module");
+		
+		// Cache Swagger Spec
+		new SwaggerSpecificationCreator().getJSON();
```
dkayiwa: Doesn't the above lead to an even slower start of openmrs? Why not do it on the first call? That way, those who will never ask for swagger docs do not get affected.

gayanW: Yeah I understand. It'll add up about 2 seconds to the startup time. However if we cache it there, load time for the docs page will be instantaneous. Please make the final call :) I'll edit then.

dkayiwa: Putting performance aside, i do not see any reason why we do not just do it the first time one tries to access the docs.

gayanW: Got it.

dkayiwa:
Can you also include a url to the ticket as advised at? https://wiki.openmrs.org/display/docs/Pull+Request+Tips

gayanW:
@dkayiwa Done.

PR #302, opened 2017-09-15
RESTWS-649 - Correct host property in OpenAPI Spec

gayanW:
* Make host and basePath not mandatory in SwaggerSpecificationCreator
* Set host and basePath to correct values in  SwaggerSpecificationController

Issue: https://issues.openmrs.org/browse/RESTWS-649

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
@@ -75,6 +75,12 @@ public class SwaggerSpecificationCreator {
 	
 	private Swagger swagger = new Swagger();
 	
+	private String host;
+	
+	private String basePath;
+	
+	private List<Scheme> schemes;
+	
 	private String baseUrl;
 	
 	private static List<SearchHandlerDoc> searchHandlerDocs;
@@ -87,10 +93,27 @@ public class SwaggerSpecificationCreator {
 	
 	private Logger log = Logger.getLogger(this.getClass());
 	
-	public SwaggerSpecificationCreator(String baseUrl) {
-		this.baseUrl = baseUrl;
-		List<SearchHandler> searchHandlers = Context.getService(RestService.class).getAllSearchHandlers();
-		searchHandlerDocs = fillSearchHandlers(searchHandlers, baseUrl);
+	public SwaggerSpecificationCreator() {
+	}
+	
+	public SwaggerSpecificationCreator host(String host) {
+		this.host = host;
+		return this;
+	}
+	
+	public SwaggerSpecificationCreator basePath(String basePath) {
+		this.basePath = basePath;
+		return this;
+	}
+	
+	public SwaggerSpecificationCreator scheme(Scheme scheme) {
+		if (schemes == null) {
+			this.schemes = new ArrayList<Scheme>();
+		}
+		if (!schemes.contains(scheme)) {
+			this.schemes.add(scheme);
+		}
+		return this;
 	}
 	
 	public String BuildJSON() {
@@ -179,7 +202,9 @@ public class SwaggerSpecificationCreator {
 		
 		this.swagger
 		        .info(info)
-		        .scheme(Scheme.HTTP)
+		        .host(this.host)
+		        .basePath(this.basePath)
+		        .schemes(this.schemes)
 		        .securityDefinition("basic_auth", new BasicAuthDefinition())
 		        .security(new SecurityRequirement().requirement("basic_auth"))
 		        .consumes("application/json")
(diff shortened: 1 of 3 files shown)


omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -806,38 +818,6 @@ private void addSubclassOperations() {
 		}
 	}
 	
-	@Deprecated
-	private List<org.openmrs.module.webservices.docs.swagger.Parameter> getParametersListForSearchHandlers(
-	        String resourceName, String searchHandlerId, int queryIndex) {
-		List<org.openmrs.module.webservices.docs.swagger.Parameter> parameters = new ArrayList<org.openmrs.module.webservices.docs.swagger.Parameter>();
-		String resourceURL = getResourceUrl(getBaseUrl(), resourceName);
-		for (SearchHandlerDoc searchDoc : searchHandlerDocs) {
```
dkayiwa: Is removing this block of code related to? "Correct host property in OpenAPI Spec"

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -87,10 +86,27 @@
 	
 	private Logger log = Logger.getLogger(this.getClass());
 	
-	public SwaggerSpecificationCreator(String baseUrl) {
-		this.baseUrl = baseUrl;
```
dkayiwa: Is removing the lines below related to? "Correct host property in OpenAPI Spec"

dkayiwa:
@gayanW this pull request seems to have more changes that needed for just "Correct host property in OpenAPI Spec". Can you create a separate pull request for other things instead of including them all in this?

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
gayanW: We no longer use the baseUrl constructor. However the `searchHandlerDocs` instance variable requires baseUrl param to initiate. `searchHandlerDocs` is only referenced by getParametersListForSearchHandlers() which is unused. So it makes sense to remove these here.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
gayanW: See the comment above.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
dkayiwa: I do not refute the fact that we no longer need some code. What am simply saying is that if removing it is not related to "Correct host property in OpenAPI Spec", then it should be in a separate pull request.

gayanW:
Please see now. baseUrl was being used to generate the host, basePath, and scheme. Yet the logic there was invalid. So it needs to be removed.

Now in the controller we pass the accurate values.

dkayiwa:
@gayanW do you mind squashing the commits?

gayanW:
Done.

omod/src/main/java/org/openmrs/module/webservices/rest/web/controller/SwaggerSpecificationController.java
```diff
@@ -27,45 +27,12 @@
...
 	String getSwaggerSpecification(HttpServletRequest request) throws Exception {
 		
-		String swaggerSpecificationJSON = "";
-		StringBuilder baseUrl = new StringBuilder();
-		String scheme = request.getScheme();
-		int port = request.getServerPort();
-		
-		baseUrl.append(scheme); // http, https
-		baseUrl.append("://");
-		baseUrl.append(request.getServerName());
```
dkayiwa: Does this ticket involve removing all these lines of code (from line 30 - 68)?
Are you sure we are not having any regressions here?

omod/src/main/java/org/openmrs/module/webservices/rest/web/controller/SwaggerSpecificationController.java
```diff
@@ -27,45 +27,12 @@
...
-		String urlWithoutScheme = "";
-		
-		/* Swagger appends scheme to urls, so we should remove it */
-		if (scheme.equals("http"))
-			urlWithoutScheme = resourcesUrl.replace("http://", "");
-		
-		else if (scheme.equals("https"))
-			urlWithoutScheme = resourcesUrl.replace("https://", "");
-		
-		SwaggerSpecificationCreator creator = new SwaggerSpecificationCreator(urlWithoutScheme);
```
gayanW: All this logic does nothing but build the baseUrl (urlWithoutScheme) string and then pass it to the SwaggerSpecificationCreator(..). As noted before generated string was invalid. Thus results in invalid host, and basePath in the spec. No there's no regression.

PR #313, opened 2017-12-07
RESTWS-681: Document TaskActionResource

Sub6Resources:
I have documented TaskActionResource with a getCREATEModel method.

See https://issues.openmrs.org/browse/RESTWS-681


dkayiwa:
You need to override three methods instead of one, as stated on the ticket.

Sub6Resources:
I mistakenly put the getCREATEModel method in the wrong file. I am closing this pull request and will open another rather than deal with revising the commit.

PR #328, opened 2018-03-02
RESTWS-382 Add Access-Control-Allow-Origin in response headers

gayanW:
Issue: [RESTWS-382](https://issues.openmrs.org/browse/RESTWS-382)

Since we have authentication in place it is should be okay for the browsers to be able to make cross-origin requests to the API. APIs such as Jira, and GitHub does this.

The test I added here in webserveices.rest-omod-common fails when calling `handle(newGetRequest(uri));`. I think I've missed some initial configuration that requires for this test. Please have a look. I'll fix it in the next commit.

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
@@ -1,0 +1,45 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.filter;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * Generic request filter intended for all /ws/rest calls <br/>
+ */
+public class RequestFilter implements Filter {
+	
+	protected final Log log = LogFactory.getLog(getClass());
+	
+	@Override
+	public void init(FilterConfig filterConfig) throws ServletException {
+		log.debug("Initializing REST WS generic request filter");
+	}
+	
+	@Override
+	public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
+	        throws IOException, ServletException {
+		HttpServletResponse response = (HttpServletResponse) servletResponse;
+		response.addHeader("Access-Control-Allow-Origin", "*");
+		
+		// continue with the filter chain in all circumstances
+		filterChain.doFilter(servletRequest, response);
+	}
+	
+	@Override
+	public void destroy() {
+		log.debug("Destroying REST WS generic request filter");
+	}
+}
--- a/omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/RequestFilterTest.java
+++ b/omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/RequestFilterTest.java
@@ -1,0 +1,43 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.filter;
+
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.webservices.rest.web.v1_0.controller.RestControllerTestUtils;
+import org.springframework.mock.web.MockHttpServletResponse;
+
+public class RequestFilterTest extends RestControllerTestUtils {
+	
+	@Before
+	public void setUp() {
+	}
+	
+	@Test
+	public void testCORS() throws Exception {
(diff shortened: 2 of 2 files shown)


omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/RequestFilterTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.mock.web.MockHttpServletResponse;
+
+public class RequestFilterTest extends RestControllerTestUtils {
+	
+	@Before
+	public void setUp() {
+	}
+	
+	@Test
+	public void testCORS() throws Exception {
```
gayanW: Stacktrace for the test RequestFilterTest:testCORS
https://pastebin.com/tvyCaitA

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/RequestFilterTest.java
```diff
@@ -0,0 +1,43 @@
...
+		String uri = getURI() + "/" + getUuid();
+		MockHttpServletResponse response = handle(newGetRequest(uri));
+		Assert.assertTrue(response.containsHeader("Access-Control-Allow-Origin"));
```
gayanW: Stacktrace for the test RequestFilterTest:testCORS
https://pastebin.com/tvyCaitA

wluyima:
Thanks @gayanW , some comments below:

- Where is the ticket associated to this pull request?
- I think theres is some more things the code needs to address, you need to limit the access to specific hosts, I believe there is a GP for limiting which host IP addresses should be granted access, your code can reuse the same GP.
- Processing of the request should not proceed for a preflight request.
- The admin should be able to turn on/off this feature with some sort of GP, by default it should be turned off.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
```diff
@@ -0,0 +1,45 @@
...
+import org.apache.commons.logging.LogFactory;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * Generic request filter intended for all /ws/rest calls <br/>
+ */
+public class RequestFilter implements Filter {
```
wluyima: Rename this to something more intuitive like CrossOriginResourceSharingFilter or even just CorsFilter may be better.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import javax.servlet.*;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * Generic request filter intended for all /ws/rest calls <br/>
```
wluyima: Javadoc need improvement

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.filter;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import javax.servlet.*;
```
wluyima: By convention, we discourage wildcard imports

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
```diff
@@ -0,0 +1,45 @@
...
+		log.debug("Initializing REST WS generic request filter");
+	}
+	
+	@Override
+	public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse, FilterChain filterChain)
+	        throws IOException, ServletException {
+		HttpServletResponse response = (HttpServletResponse) servletResponse;
+		response.addHeader("Access-Control-Allow-Origin", "*");
+		
+		// continue with the filter chain in all circumstances
```
wluyima: You should not proceed for a preflight request

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/RequestFilter.java
wluyima: I don't see where you have registered the filter

gayanW: Noted.

gayanW:
Thanks for the input. This actually is a test PR. The issue have been abandoned for sometime now, as it is not a must. But I thought of giving it another go.

    Where is the ticket associated to this pull request?

Updated.

    I think theres is some more things the code needs to address, you need to limit the access to specific hosts, I believe there is a GP for limiting which host IP addresses should be granted access, your code can reuse the same GP.

Yeah, I've seen it. Since they are actually is two separate cases wouldn't it be better if we keep these logics separate? 

Say an IP is whitelisted, so returning CORS header should be a must, for that client IP to be able to make requests to the the server from a browser.
Say an IP is not whitelisted. Even in here returning CORS header doesn't do any harm, **if we register CORS filter after the AuthorizationFilter.**

So adding CORS header, after the [AuthorizationFiler](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/AuthorizationFilter.java) will be enough, right?

    Processing of the request should not proceed for a preflight request.
    The admin should be able to turn on/off this feature with some sort of GP, by default it should be turned off.

Sorry I don't understand the preflight request thing.  Isn't it issued by the browser?
Yeah, that really makes sense to  add a toggle on/off feature. I'll do the suggested changes.

wluyima:
The user's session is usable in the browser by other external scrips to make calls, so because someone is authenticated doesn't mean the calls are from a trusted source, they could be from an attacker since they have access to the sessionId inside the browser.

gayanW:
> The user's session is usable in the browser by other external scrips to make calls, so because someone is authenticated doesn't mean the calls are from a trusted source, they could be from an attacker since they have access to the sessionId inside the browser.

Okay. So what you suggest is that we create separate CORS filter (to allow specific hosts) and it's respective admin ui option as similar to the current IP filtering feature?

Looking forward to your reply, so I could proceed. Thanks.

wluyima:
Something along those lines

dkayiwa:
@gayanW did you see the travis failure?

gayanW:
@dkayiwa I'll not be available for few months. I'll create a new PR, when I get the time. Thanks.

PR #333, opened 2018-05-18
RESTWS-706 Add sorting by dateActivated Or dateStopped to Orders

fanyui:
Ticket description found [RESTWS-706](https://issues.openmrs.org/browse/RESTWS-706)

### DIFF
--- a/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
+++ b/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
@@ -11,6 +11,9 @@ package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_10;
 
 import java.util.Date;
 import java.util.List;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 
 import io.swagger.models.Model;
 import io.swagger.models.ModelImpl;
@@ -39,6 +42,7 @@ import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOp
 import org.openmrs.module.webservices.rest.web.response.ResponseException;
 import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.OrderResource1_8;
 import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PatientResource1_8;
+import org.openmrs.util.OpenmrsUtil;
 
 /**
  * {@link org.openmrs.module.webservices.rest.web.annotation.Resource} for {@link org.openmrs.Order}
@@ -226,6 +230,8 @@ public class OrderResource1_10 extends OrderResource1_8 {
 			
 			String careSettingUuid = context.getRequest().getParameter("careSetting");
 			String asOfDateString = context.getRequest().getParameter("asOfDate");
+			String sortParam = context.getRequest().getParameter("sort");
+			
 			CareSetting careSetting = null;
 			Date asOfDate = null;
 			if (StringUtils.isNotBlank(asOfDateString)) {
@@ -243,10 +249,37 @@ public class OrderResource1_10 extends OrderResource1_8 {
 				filterByType(orders, context.getType());
 			}
 			
-			return new NeedsPaging<Order>(orders, context);
+			if (StringUtils.isNotBlank(sortParam)) {
+				List<Order> sortedOrder = sortOrdersBasedOnDateActivatedOrDateStopped(orders, sortParam, status);
+				return new NeedsPaging<Order>(sortedOrder, context);
+			}
+			else {
+				return new NeedsPaging<Order>(orders, context);
+			}
 		}
 		
 		return new EmptySearchResult();
+	}
+	
+	public  List<Order> sortOrdersBasedOnDateActivatedOrDateStopped(List<Order> orders, final String sortOrder,
+	        final String status) {
+		List<Order> sortedList = new ArrayList<Order>(orders);
+		
+		Collections.sort(sortedList, new Comparator<Order>() {
+			
+			@Override
+			public int compare(Order o1, Order o2) {
+				if (status.equalsIgnoreCase("inactive")) {
+					return sortOrder.equalsIgnoreCase("ASC") ? o1.getDateStopped().compareTo(o2.getDateStopped()) : o2
+					        .getDateStopped().compareTo(o1.getDateStopped());
+				}
+				else {
+					return sortOrder.equalsIgnoreCase("asc") ? o1.getDateActivated().compareTo(o2.getDateActivated()) : o2
+					        .getDateActivated().compareTo(o1.getDateActivated());
+				}
+			}
+		});
+		return sortedList;
 	}
 	
 	/**
--- a/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
+++ b/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
@@ -260,22 +260,28 @@ public class OrderResource1_10 extends OrderResource1_8 {
 		
 		return new EmptySearchResult();
 	}
+
+	private static Date getUsableDate(Order order) {
+		return order.getDateStopped() != null ? order.getDateStopped() : (order.getAutoExpireDate() != null ? order.getAutoExpireDate() : order.getDateCreated());
+	}
 	
-	public  List<Order> sortOrdersBasedOnDateActivatedOrDateStopped(List<Order> orders, final String sortOrder,
+	private static Date getActiveOrderSortDate(Order order) {
+		return order.getDateActivated() != null ? order.getDateActivated() : order.getDateCreated();	
+	}
+	public List<Order> sortOrdersBasedOnDateActivatedOrDateStopped(List<Order> orders, final String sortOrder,
 	        final String status) {
 		List<Order> sortedList = new ArrayList<Order>(orders);
-		
 		Collections.sort(sortedList, new Comparator<Order>() {
 			
 			@Override
 			public int compare(Order o1, Order o2) {
 				if (status.equalsIgnoreCase("inactive")) {
-					return sortOrder.equalsIgnoreCase("ASC") ? o1.getDateStopped().compareTo(o2.getDateStopped()) : o2
-					        .getDateStopped().compareTo(o1.getDateStopped());
+					return sortOrder.equalsIgnoreCase("ASC") ?
+					        getUsableDate(o1).compareTo(getUsableDate(o2)) : getUsableDate(o2).compareTo(getUsableDate(o1));
 				}
 				else {
-					return sortOrder.equalsIgnoreCase("asc") ? o1.getDateActivated().compareTo(o2.getDateActivated()) : o2
-					        .getDateActivated().compareTo(o1.getDateActivated());
+							
+					return sortOrder.equalsIgnoreCase("asc") ? getActiveOrderSortDate(o1).compareTo(getActiveOrderSortDate(o2)) : getActiveOrderSortDate(o2).compareTo(getActiveOrderSortDate(o1));
 				}
 			}
 		});
--- a/omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/OrderController1_10Test.java
+++ b/omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/OrderController1_10Test.java
@@ -388,9 +388,7 @@ public class OrderController1_10Test extends MainResourceControllerTest {
 		        PropertyUtils.getProperty(resultList.get(2), "uuid").toString(),
 		        PropertyUtils.getProperty(resultList.get(3), "uuid").toString(),
 		        PropertyUtils.getProperty(resultList.get(4), "uuid").toString() });
-		assertThat(uuids, hasItems(expectedOrderUuids));
-	
-assertThat(uuids, contains( orderService.getOrder(7).getUuid(), orderService.getOrder(5).getUuid(), orderService.getOrder(444).getUuid(), orderService.getOrder(3).getUuid(), orderService.getOrder(222).getUuid()));
+		assertThat(uuids, contains( orderService.getOrder(7).getUuid(), orderService.getOrder(5).getUuid(), orderService.getOrder(444).getUuid(), orderService.getOrder(3).getUuid(), orderService.getOrder(222).getUuid()));
 	}
 	
 	@Test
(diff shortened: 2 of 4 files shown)


mogoodrich:
Generally looks good, @fanyui but I'd recommend a few changes...

Can you add some unit tests?  I think you just should be able to start with the following test and add the "sort" parameter:

https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/OrderController1_10Test.java#L418

Small point would be to use "equalsIgnoreCase" instead of "compareToIgnoreCase".

Also, I will be on vacation for a week, starting tomorrow, but if you flag @djazayeri  or @mseaton  they could quickly review once you've made your changes.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
```diff
@@ -243,12 +248,52 @@ protected PageableResult doSearch(RequestContext context) {
...
+					}
+					else {
+						return o2.getDateActivated().compareTo(o1.getDateActivated());
+					}
+					
+				}
+			});
+		}
+	}
+
```
larrystone: I would suggest simplifying lines 263 - 295 further
```
Collections.sort(orders, new Comparator<Order>() {		
				@Override
				public int compare(Order o1, Order o2) {
					if (status.compareToIgnoreCase("inactive") == 0) {
						return sortOrder.compareToIgnoreCase("ASC") == 0 ? o1.getDateStopped().compareTo(o2.getDateStopped()) : o2.getDateStopped().compareTo(o1.getDateStopped());
					}
					else {
						return sortOrder.compareToIgnoreCase("ASC") == 0 ? o1.getDateActivated().compareTo(o2.getDateActivated()) : o2.getDateActivated().compareTo(o1.getDateActivated());
					}
				}
			});
```
I am just worried about code duplication. I may be wrong though

mogoodrich: I'd agree with this change a well.  Thanks @larrystone !

larrystone:
Hi @fanyui How is this project coming along? Here's a gentle reminder that some features of OrderEntry UI depends on it and don't hesitate to reach out if you are having any challenge. Thanks
cc @dkayiwa @djazayeri

dkayiwa:
@fanyui can i look at the failure log?

larrystone:
Is this happening after you have written test for this feature or is it that the existing test was failing after making the above changes?

fanyui:
@larrystone @dkayiwa   it is failing because am trying to add a test for it. existing tests have no problem. 
Here is what i did. Manually looked at the dateActivated and Uuid for the various orders used in the test and sorted them. then i compare with the result of the sorted one done by the system.
Error expected {set of expected ids } in sorted order but found {set of ids still not sorted} this happens only for shouldGetTheActiveDrugOrdersForAPatient  but has no problem with shouldGetTheActiveOrdersForAPatient and looking at the doSearch method the shouldGetTheActiveDrugOrdersForAPatient having the drugorder in it url does not get the sort method applied to it. I wish to know how i can get that fixed

dkayiwa:
Commit the failing test and i take a look.

fanyui:
@dkayiwa  I have added the commit with the failing test.

larrystone:
Okay, I will pull this changes to my local machine, test it out and give you feedback tomorrow. Nice work man

dkayiwa:
@fanyui that one goes through DrugOrderSubclassHandler1_10

fanyui:
Thanks @dkayiwa  @larrystone  I have updated the pull request

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
```diff
@@ -243,12 +248,34 @@ protected PageableResult doSearch(RequestContext context) {
...
+				sortOrdersBasedOnDateActivatedOrDateStopped(orders, sortParam, status);
+			}
+			
 			return new NeedsPaging<Order>(orders, context);
 		}
 		
 		return new EmptySearchResult();
 	}
 	
+	public static void sortOrdersBasedOnDateActivatedOrDateStopped(List<Order> orders, final String sortOrder,
```
dkayiwa: I think it will be better if the method returns the sorted list.

fuchodeveloper:
Hi @fanyui 
Please have you been able to test your implementation with the local OpenMRS instance?

larrystone:
@fanyui We just tested the omod on our local orderEntry UI implementation and it works. We are still testing but I guess it's looking good. Good job man 👍

fanyui:
@larrystone  Thanks Happy to have this task completed with your help and the entire community.

fuchodeveloper:
Hi @fanyui 
I am currently working on the frontend implementation for sorting orders that require making use of your work. The challenge I face occurs when a request is made to return past orders that auto-expired.  The reason for this issue is because the current implementation does not handle situations where a past order was auto-expired not stopped. What this means in this situation is that the `dateStopped` field which is what the sorting is based on is `null`, but `autoExpireDate` is a valid date showing when the order expired.

One way to resolve this would be to support sorting of past orders based on `dateStopped` or `autoExpireDate`.

cc: @dkayiwa

mogoodrich:
Hmm... good point @fuchodeveloper... I think it's also possible (maybe not in the Order Entry UI) to have a order with a "scheduleDate" but not a "stoppedDate", "autoExpiredDate" or a "dateActivated"?

I'm thinking for simplicity we just determine the date to sort on based on dateActivated or dateCreated (since dateCreated *must* exist) regardless of whether it's an active or inactive order.

So in the comparator you'd do something like:

o1SortDate = o1.getDateActivated() != null ? o1.getDateActiviated() : o1.getDateCreated()
o2SortDate = o2.getDateActivated() != null ? o2.getDateActiviated() : o2.getDateCreated()

And then "o1SortDate.compareTo(o2SortDate)"  if ascending and reverse if descending.

Does this make sense? Do we have a use case in the UI were we really want to sort by dateStopped or autoexpireDate? 

 fyi @dkayiwa  @djazayeri

larrystone:
@mogoodrich Thanks for the insight. However, one of the challenges with using using dateActivated for inactive orders is that the records being displayed may not look sorted. For example consider this data
```
[{"dateActivated" : "22/5/2016", "dateStopped" : "26/5/2016", ... },
{"dateActivated" : "23/5/2016", "dateStopped" : "23/5/2016", ... },
{"dateActivated" : "24/5/2016", "dateStopped" : "25/5/2016", ... }]
```
When we try to display this on the inactive orders section on the UI, we expect that the records be displayed according to the dates which they were stopped because we should be able to see the recently stopped orders and not based on dateActivated like this:
```
[{"dateActivated" : "23/5/2016", "dateStopped" : "23/5/2016", ... },
{"dateActivated" : "24/5/2016", "dateStopped" : "25/5/2016", ... },
{"dateActivated" : "22/5/2016", "dateStopped" : "26/5/2016", ... }]
```
What do you think?

fanyui:
@fuchodeveloper nice catch.  I think your suggestion is a good one if there actually is a necessity to get autoexpired orders. could you give an example of the search query you issued?

omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/OrderController1_10Test.java
```diff
@@ -420,14 +423,15 @@ public void shouldGetTheActiveDrugOrdersForAPatient() throws Exception {
...
-		        "da7f524f-27ce-4bb2-86d6-6d1d05312bd5"))));
+		        "da7f524f-27ce-4bb2-86d6-6d1d05312bd5"), new Parameter("sort", "desc"), new Parameter("status", "active"))));
 		assertEquals(expectedOrderUuids.length, Util.getResultsSize(results));
 		List<Object> resultList = Util.getResultsList(results);
 		List<String> uuids = Arrays.asList(new String[] { PropertyUtils.getProperty(resultList.get(0), "uuid").toString(),
 		        PropertyUtils.getProperty(resultList.get(1), "uuid").toString(),
 		        PropertyUtils.getProperty(resultList.get(2), "uuid").toString(),
 		        PropertyUtils.getProperty(resultList.get(3), "uuid").toString(), });
 		assertThat(uuids, hasItems(expectedOrderUuids));
+assertThat(uuids, contains(orderService.getOrder(5).getUuid(), orderService.getOrder(444).getUuid(), orderService.getOrder(3).getUuid(), orderService.getOrder(222).getUuid()));
```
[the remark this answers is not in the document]
fanyui: Noted I will address fix those

larrystone:
@fanyui Okay, here is the challenge we had, if an order when not explicitly stopped but was left to autoExpire then the dateStopped field will be null (Yes, there is a logic that auto handles autoExpiration of Orders). This resulted in a 500 error because your implementation above was making use of dateStopped which is not always guaranteed to be available

fuchodeveloper:
@fanyui 
Yes, getting auto-expired orders could be necessary in situations where the order was not explicitly stopped but was allowed to expire. The query would remain the same as illustrated below:
`/order?totalCount=true&sort=desc&status=inactive&limit=${limit}&startIndex=${startIndex}&careSetting=${careSetting}&patient=${patientUuid}&t=drugorder&v=full`

mogoodrich:
@larrystone  @fanyui 

We can't handle sorting on the client-side because of server pagination, correct?

fanyui:
@larrystone  What if we take care of the sorting by using nulls for inactive orders as latest then surely the 500 error will no longer exist but. @fuchodeveloper  That is assuming no one explicitely requested for auto expired orders. @fuchodeveloper  I was thinking maybe you added a parameter to the url to only get auto-expired orders but it doesn't seem to be the case

fuchodeveloper:
@fanyui 
No, I did not add any new parameters. To give a bit more context to the situation at hand.
Say we have one active order for `Paracetamol`. If the `dateActivated` is 29-05-2018 08:20 and the drug order is set for the duration of two days. Which means `Paracetamol` would auto-expire on 31-05-2018 08:20.

If for some reason `Paracetamol` is stopped on 30th, the `dateStopped` value could be, for instance,  30-05-2018 16:00. However, if the drug truly expires on the 31th as prescribed, then `dateStopped` value would be `null` because it was not stopped before it expired. This is what causes the `500` error to be returned when a request is made to sort and fetch the past drug orders.
Hope this was helpful.

cc: @dkayiwa

djazayeri:
Probably ignore this comment given where you're at in the dev cycle for this ticket, but just to comment offhand:

1. the best thing is probably to sort on "the last date of any change", which could be activation, scheduled start, expiration, stoppage.
2. although this sorting is happening in java code now, I would hope it eventually gets moved to the db. So ideally whatever sorting is done here is simple enough that it can be done with a clause like `order by nvl(o.date_stopped, o.auto_expire_date)`

mogoodrich:
I would support either the solution @darius suggested (sorting on the "last date of any change") or the solution I proposed above it.

fuchodeveloper:
Thank you @djazayeri and @mogoodrich for your feedback.

@fanyui please, could you update the pull request with the suggestions?
Kindly reach out if you have any blockers.

_NB:_ [OEUI-101](https://issues.openmrs.org/browse/OEUI-101) is dependent on the output from this implementation.

fanyui:
Thanks everyone I will go with @mogoodrich  suggestion for now. But as @djazayeri  mentioned i agree with him on moving all these sorting to the db.

fuchodeveloper:
@fanyui 
Worked with @larrystone to come up with the solution below:
```
	private static Date getUsableDate(Order order) {
		return order.getDateStopped() != null ? order.getDateStopped() : order.getAutoExpireDate();
	}
	
	public static void sortOrdersBasedOnDateActivatedOrDateStopped(List<Order> orders, final String sortOrder,
	        final String status) {
		Collections.sort(orders, new Comparator<Order>() {
			
			@Override
			public int compare(Order o1, Order o2) {
				if (status.equalsIgnoreCase("inactive")) {
					return sortOrder.equalsIgnoreCase("ASC") ?
					        getUsableDate(o1).compareTo(getUsableDate(o2)) : getUsableDate(o2).compareTo(getUsableDate(o1));
				}
				else {
					return sortOrder.equalsIgnoreCase("asc") ? o1.getDateActivated().compareTo(o2.getDateActivated()) : o2
					        .getDateActivated().compareTo(o1.getDateActivated());
				}
			}
		});
	}
```

Have tested it with the local instance of the Order Entry UI OWA and it works as expected.

cc: @dkayiwa @mogoodrich @djazayeri

fanyui:
@fuchodeveloper clean code i had made a commit here https://github.com/openmrs/openmrs-module-webservices.rest/pull/333/commits/15edc847d1beff8dcb02e2984a01f5c156471129 and was still testing on my local instance. I'll update and test with yours as well.

fanyui:
@larrystone  @fuchodeveloper I've updated and tested with the code you gave everything seems fine on my localhost. I will like to get how it behaves in your local instances. @mogoodrich  the constains method from  org.hamcrest.Matchers.contains; doesnot take list or even arrays so it required me to have use matcher items of convert the expectedOrderUuid before passing it to the method which i saw as an overhead so i just removed the hasItems and used contains but still with the items passed to it.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
```diff
@@ -243,11 +249,41 @@ protected PageableResult doSearch(RequestContext context) {
...
+			else {
+				return new NeedsPaging<Order>(orders, context);
+			}
 		}
 		
 		return new EmptySearchResult();
 	}
+
+	private static Date getUsableDate(Order order) {
+		return order.getDateStopped() != null ? order.getDateStopped() : order.getAutoExpireDate();
```
mogoodrich: We should have a final fallback here if dateStopped and getAutoExpireDate are both null... (not sure if that should ever happen, but both fields are allowed to be null in the db so it's *possible* it could happen).  I would just fall back to dateCreated (since it is the one date that is guaranteed never to be null).

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
```diff
@@ -243,11 +249,41 @@ protected PageableResult doSearch(RequestContext context) {
...
+		Collections.sort(sortedList, new Comparator<Order>() {
+			
+			@Override
+			public int compare(Order o1, Order o2) {
+				if (status.equalsIgnoreCase("inactive")) {
+					return sortOrder.equalsIgnoreCase("ASC") ?
+					        getUsableDate(o1).compareTo(getUsableDate(o2)) : getUsableDate(o2).compareTo(getUsableDate(o1));
+				}
+				else {
+					return sortOrder.equalsIgnoreCase("asc") ? o1.getDateActivated().compareTo(o2.getDateActivated()) : o2
```
mogoodrich: We should have a fallback to dateCreated here as well... I believe that this branch of the if/else will be hit if status is set to "active" or "any"  (see https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderUtil.java#L40)  so it's possible that we would get orders here that have dateActivated = null.

dkayiwa:
@fanyui are you still working on this?

mogoodrich:
Sorry, @fanyui, I had missed that you had made a further update... go ahead and flag me on the pull request whenever you have something new for me to review!

This looks good!  I will merge this in now, thanks so much!

PR #335, opened 2018-08-22
RESTWS-711- Change to new MappingJackson2HttpMessageConverter in spring 4.3.x

wluyima:
https://issues.openmrs.org/browse/RESTWS-711


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/MappingJacksonHttpMessageConverter.java
```diff
@@ -0,0 +1,40 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web;
+
+import org.openmrs.api.context.Context;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.converter.HttpMessageConverter;
+
+@Configuration
```
djazayeri: Need some javadoc explaining what's going on here.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/MappingJacksonHttpMessageConverter.java
```diff
@@ -0,0 +1,40 @@
...
+public class MappingJacksonHttpMessageConverter {
+	
+	private static final String PACKAGE = "org.springframework.http.converter.json.";
+	
+	private static final String OLD_CLASS_NAME = PACKAGE + "MappingJacksonHttpMessageConverter";
```
djazayeri: Instead of OLD_CLASS_NAME and NEW_CLASS_NAME wouldn't it be better to call these CLASS_FROM_SPRING43 and CLASS_BEFORE_SPRING43 or something?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/MappingJacksonHttpMessageConverter.java
wluyima: Will do, actually I need to rename this class

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/MappingJacksonHttpMessageConverter.java
wluyima: Makes sense

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/MappingJacksonHttpMessageConverterConfig.java
```diff
@@ -0,0 +1,47 @@
...
+ * configuration class allows the module to run on both pre and post 4.1.0 versions of Spring by
+ * loading the new class if it's available in the spring version that OpenMRS is running on
+ * otherwise it falls back to the old one.
+ */
+@Configuration
+public class MappingJacksonHttpMessageConverterConfig {
+	
+	private static final String PACKAGE = "org.springframework.http.converter.json.";
+	
+	private static final String CLASS_BEFORE_SPRING_4_1 = PACKAGE + "MappingJacksonHttpMessageConverter";
```
wluyima: Actually it's more like the OpenMRS version rather than the spring version. When we upgraded to spring 4.1.x, the dev copied the removed classes into core because I guess we were still running Jackson 1.x yet the new converter requires Jackson 2.x

PR #336, opened 2018-08-24
Solving RESTWS-713 - Add free-text cause of death to Person resource

fruether:
1. Adding new attribute to set of return values
2. Create one simple test case to make sure everything is working

https://issues.openmrs.org/browse/RESTWS-713

### DIFF
--- a/integration-tests/src/test/java/org/openmrs/module/webservices/rest/ITBase.java
+++ b/integration-tests/src/test/java/org/openmrs/module/webservices/rest/ITBase.java
@@ -64,11 +64,11 @@ public abstract class ITBase {
 				
 				final RequestConfig requestConfig = RequestConfig.custom().setSocketTimeout(retryAfter)
 				        .setConnectTimeout(retryAfter).build();
-
+				
 				final String startupUri = TEST_URL.getScheme() + "://" + TEST_URL.getHost() + ":"
-						+ TEST_URL.getPort() + TEST_URL.getPath();
+				        + TEST_URL.getPort() + TEST_URL.getPath();
 				System.out.println("Waiting for server at " + startupUri + " for " + timeout / 1000 + " more seconds...");
-
+				
 				while (System.currentTimeMillis() - time < timeout) {
 					try {
 						final HttpClient client = HttpClientBuilder.create().disableAutomaticRetries().build();
@@ -97,10 +97,10 @@ public abstract class ITBase {
 					catch (IOException e) {
 						System.out.println(e.toString());
 					}
-
+					
 					try {
 						System.out.println("Waiting for "
-								+ (timeout - (System.currentTimeMillis() - time)) / 1000 + " more seconds...");
+						        + (timeout - (System.currentTimeMillis() - time)) / 1000 + " more seconds...");
 						Thread.sleep(retryAfter);
 					}
 					catch (InterruptedException e) {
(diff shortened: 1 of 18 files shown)


omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
```diff
@@ -0,0 +1,29 @@
...
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_2;
+
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PersonResource1_8;
+
+public class PersonResource2_2 extends PersonResource1_8 {
```
wluyima: Should extend the current highest class in the hierarchy which I think was suggested to be `PersonResource1_11` from the ticket description.

There is missing metadata on the class for this to work, please refer to `PersonResource1_11` for an example

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2Test.java
```diff
@@ -0,0 +1,58 @@
...
+	
+	@Override
+	public String getUuidProperty() {
+		return RestTestConstants2_2.PERSON_UUID;
+	}
+	
+	@Override
+	public void validateFullRepresentation() throws Exception {
+		super.validateFullRepresentation();
+		assertPropEquals("gender", getObject().getGender());
```
wluyima: This is repetition, the call to super.validateFullRepresentation() avove has already validated all the existing fields, therefore you should only validate the new one after that.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
```diff
@@ -0,0 +1,29 @@
...
+	
+	@Override
+	public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+		DelegatingResourceDescription description = super.getRepresentationDescription(rep);
+		
+		if(description != null) {
+			description.addProperty("causeOfDeathNonCoded");
```
wluyima: I think you should check the representation type so that you don't add the field for the RefRepresentation

fruether: But it is the same way done in 1_11. There is is also just checked if it is != null.

Shall I yet do the instance of check?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
fruether: oh sry. For some reasons I had 1_8 in my mind while creating the class

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
wluyima: It was just an example to follow, you shouldn't copy everything literally, context and the property you're adding determines, in this case causeOfDeath is null in most cases for most patients, therefore it's not key information, so you should not include it for the RefRepresentation which  is a very minimal form.

fruether:
Alright I fixed this in the new PR version.

PR #337, opened 2018-08-24
Solving RESTWS-713

fruether:
Adding new attribute to set of return values
Create one simple test case to make sure everything is working
https://issues.openmrs.org/browse/RESTWS-713

### DIFF
--- a/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
+++ b/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
@@ -1,0 +1,64 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+
+package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_2;
+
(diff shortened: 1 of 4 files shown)


fruether:
Would it be possible to change the reference Person object so that it contains actually a causeofDeath non coded ?

fruether:
I have to add methods like getCreatableProperties (if I understood the example correctly) so the PR is not ready yet

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/PersonResource2_2.java
```diff
@@ -0,0 +1,64 @@
...
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingCrudResource;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_11.PersonResource1_11;
+
+/**
+ * {@link Resource} for Person, supporting standard CRUD operations
+ */
+@Resource(name = RestConstants.VERSION_3 + "/person", order = 1, supportedClass = Person.class, supportedOpenmrsVersions = {
```
wluyima: I doubt if order adds value here? Given that it's the same value anyways like PersonResource1_11

wluyima: Should be version 1, this is rest API and not a resource version.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/RestConstants.java
```diff
@@ -73,6 +73,11 @@
 	 */
 	public static final String VERSION_2 = "v2";
 	
+	/**
+	 * The version number for the third rest web services representations
+	 */
+	public static final String VERSION_3 = "v3";
```
wluyima: You can remove this

PR #340, opened 2018-08-28
Solving RESTWS-714 - Add fulfillerStatus and Orders.fulfillerComment to Order resource

fruether:
This PR is not yet working but I added it to give a discussion base for the Talk thread: 

The error is basically is the following message: 

`org.springframework.http.converter.HttpMessageNotWritableException: Could not write JSON: Direct self-reference leading to cycle (through reference chain: org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_2.FulfillerDetailsRepresentation["parent"]->org.openmrs.TestOrder["creator"]->org.openmrs.User_$$_jvstbb0_55["creator"]); nested exception is org.codehaus.jackson.map.JsonMappingException: Direct self-reference leading to cycle `

### DIFF
--- a/integration-tests/src/test/java/org/openmrs/module/webservices/rest/ITBase.java
+++ b/integration-tests/src/test/java/org/openmrs/module/webservices/rest/ITBase.java
@@ -64,11 +64,11 @@ public abstract class ITBase {
 				
 				final RequestConfig requestConfig = RequestConfig.custom().setSocketTimeout(retryAfter)
 				        .setConnectTimeout(retryAfter).build();
-
+				
 				final String startupUri = TEST_URL.getScheme() + "://" + TEST_URL.getHost() + ":"
-						+ TEST_URL.getPort() + TEST_URL.getPath();
+				        + TEST_URL.getPort() + TEST_URL.getPath();
 				System.out.println("Waiting for server at " + startupUri + " for " + timeout / 1000 + " more seconds...");
-
+				
 				while (System.currentTimeMillis() - time < timeout) {
 					try {
 						final HttpClient client = HttpClientBuilder.create().disableAutomaticRetries().build();
@@ -97,10 +97,10 @@ public abstract class ITBase {
 					catch (IOException e) {
 						System.out.println(e.toString());
 					}
-
+					
 					try {
 						System.out.println("Waiting for "
-								+ (timeout - (System.currentTimeMillis() - time)) / 1000 + " more seconds...");
+						        + (timeout - (System.currentTimeMillis() - time)) / 1000 + " more seconds...");
 						Thread.sleep(retryAfter);
 					}
 					catch (InterruptedException e) {
--- a/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/DrugOrderSubclassHandler1_10.java
+++ b/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/DrugOrderSubclassHandler1_10.java
@@ -173,14 +173,15 @@ public class DrugOrderSubclassHandler1_10 extends DrugOrderSubclassHandler1_8 {
 		OrderResource1_10 orderResource = (OrderResource1_10) Context.getService(RestService.class)
 		        .getResourceBySupportedClass(Order.class);
 		if (StringUtils.isNotBlank(sortParam)) {
-			List<Order> sortedOrder = orderResource.sortOrdersBasedOnDateActivatedOrDateStopped(drugOrders, sortParam, status);
+			List<Order> sortedOrder = orderResource.sortOrdersBasedOnDateActivatedOrDateStopped(drugOrders, sortParam,
+			    status);
 			return new NeedsPaging<Order>(sortedOrder, context);
 		}
 		else {
 			return new NeedsPaging<Order>(drugOrders, context);
 		}
 	}
-
+	
 	/**
 	 * @see OrderResource1_10#getDisplayString(org.openmrs.Order)
 	 */
--- a/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
+++ b/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
@@ -260,14 +260,16 @@ public class OrderResource1_10 extends OrderResource1_8 {
 		
 		return new EmptySearchResult();
 	}
-
+	
 	private static Date getUsableDate(Order order) {
-		return order.getDateStopped() != null ? order.getDateStopped() : (order.getAutoExpireDate() != null ? order.getAutoExpireDate() : order.getDateCreated());
+		return order.getDateStopped() != null ? order.getDateStopped() : (order.getAutoExpireDate() != null ? order
+		        .getAutoExpireDate() : order.getDateCreated());
 	}
 	
 	private static Date getActiveOrderSortDate(Order order) {
-		return order.getDateActivated() != null ? order.getDateActivated() : order.getDateCreated();	
+		return order.getDateActivated() != null ? order.getDateActivated() : order.getDateCreated();
 	}
+	
 	public List<Order> sortOrdersBasedOnDateActivatedOrDateStopped(List<Order> orders, final String sortOrder,
 	        final String status) {
 		List<Order> sortedList = new ArrayList<Order>(orders);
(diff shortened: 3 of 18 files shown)


wluyima:
Typically you get that error if you have a compound field that has no representation which I think in this case is FulfillerDetailsRepresentation

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
```diff
@@ -0,0 +1,117 @@
...
+import org.openmrs.module.webservices.rest.web.resource.api.PageableResult;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingSubResource;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
+import org.openmrs.module.webservices.rest.web.response.ResponseException;
+
+/**
+ * {@link Resource} for Order, supporting standard CRUD operations on fulfiller_comment and fulfiller_status
+ */
+@SubResource(parent = OrderResource2_2.class, path = "fulfillerstatus", supportedClass = Order.class, supportedOpenmrsVersions = {
```
wluyima: I think supported class is supposed to be `FulfillerDetailsRepresentation` which should get you past the current error you getting.

fruether: oh yeah it does. Thank you!

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
```diff
@@ -0,0 +1,117 @@
...
+	
+	@Override
+	public FulfillerDetailsRepresentation newDelegate() {
+		return new FulfillerDetailsRepresentation();
+	}
+	
+	@Override
+	public FulfillerDetailsRepresentation save(FulfillerDetailsRepresentation delegate) {
+		Order order =  Context.getOrderService().updateOrderFulfillerStatus(delegate.getParent(), delegate.getFulfillerStatus(), delegate.getFulfillerComment());
+		delegate.setParent(order);
```
wluyima: Isn't this redundant? Because what value would you have got back on line before it?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
```diff
@@ -0,0 +1,117 @@
...
+	
+	@Override
+	public Model getCREATEModel(Representation rep) {
+		return new ModelImpl()
+				.property("fulfillerComment", new StringProperty())
+				.property("fulfillerStatus", new EnumProperty(null));
+	}
+	
+	@Override
+	public Model getUPDATEModel(Representation rep) {
```
wluyima: I think the resource only supports create and not update

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsRepresentation.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_2;
+
+import org.openmrs.Order;
+
+public class FulfillerDetailsRepresentation {
+	
+	private Order.FulfillerStatus fulfillerStatus;
+	private String fulfillerComment;
+	private Order parent;
```
wluyima: Can you please rename this to order for clarity?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
```diff
@@ -0,0 +1,117 @@
...
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	public FulfillerDetailsRepresentation getByUniqueId(String uniqueId) {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	protected void delete(FulfillerDetailsRepresentation delegate, String reason, RequestContext context)
```
wluyima: Do we actually want to support this?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsRepresentation.java
fruether: I am not sure what you mean? Should the order not be called parent?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
fruether: I am not sure if the created order is somehow different and hence a new parent must be set?

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2WebTest.java
```diff
@@ -0,0 +1,63 @@
...
+	@Before
+	public void setUp() throws Exception {
+		Order order = Context.getOrderService().getOrderByUuid(RestTestConstants2_2.ORDER_UUID).copy();
+		order.setUuid(RestTestConstants2_2.ORDER_UUID_Web);
+		order.setOrderId(null);
+		Context.getOrderService().saveOrder(order, null);
+	}
+	
+	@Test
+	public void shouldCreateNew() throws Exception {
```
wluyima: This test should be named along the lines of what it actually does, i.e. set fulfiller details.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
fruether: Are you sure? Because the filler details like comment and status are may updated during the order process. And I am not sure if in this scenario always the Create method would be called by the users of the the REST service. 
Or if maybe at some point some modules would Update these details by submitting it through PUT...?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
wluyima: It would be nice to have some consistency in the naming of the resource and these classes, some places you are using `FulfillerStatus` while others you are using `FulfillerDetailsRepresentation`, can you use `FulfillerDetails` or FulfillerStatus, am more inclined towards `FulfillerDetails`

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsRepresentation.java
wluyima: The field value is an Order object, therefore it's more intuitive to name it order

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
wluyima: I thought based on the logic you implemented in the OrderService.updateOrderFulfillerStatus, it updates and returns the same order object.

wluyima: You call saveOrderInternal after setting fullfiller details which returns the same order object.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerStatusResource2_2.java
wluyima: This is a unique resource because the supported class is not persistent so differentiating save vs update doesn't make much sense.

fruether:
@wluyima new PR is there: https://github.com/openmrs/openmrs-module-webservices.rest/pull/341/commits/969c2d99b789e62a1c8d6839783808bb83c64abd

I think I should have taken all the comments into account

wluyima:
FYI, you don't have to create a new PR when you make changes, just commit and push changes to your remote branch and then github will automatically update the original PR.

fruether:
Yeah it is more like happening by accident @wluyima. Sometimes I squash the commits and then somehow force an empty push to repository. That's leads to the closing of the PR :(

wluyima:
Actually the dev merging the PR will squash the commits in github, it's a feature that didn't exist at the time we wrote the documentation for PRs

PR #341, opened 2018-08-28
Solving RESTWS-714 - Add fulfillerStatus and Orders.fulfillerComment to Order resource

fruether:
Ticket:https://issues.openmrs.org/browse/RESTWS-714
Talk: https://talk.openmrs.org/t/adding-order-status-to-orders/17928/36

### DIFF
--- a/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
+++ b/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
@@ -1,0 +1,98 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_2;
+
+import io.swagger.models.Model;
+import io.swagger.models.ModelImpl;
+import io.swagger.models.properties.StringProperty;
+import org.openmrs.Order;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.webservices.docs.swagger.core.property.EnumProperty;
+import org.openmrs.module.webservices.rest.web.RequestContext;
+import org.openmrs.module.webservices.rest.web.RestConstants;
+import org.openmrs.module.webservices.rest.web.annotation.Resource;
+import org.openmrs.module.webservices.rest.web.annotation.SubResource;
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.api.PageableResult;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingSubResource;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
+import org.openmrs.module.webservices.rest.web.response.ResponseException;
+
+/**
+ * {@link Resource} for Order, supporting standard CRUD operations on fulfiller_comment and
+ * fulfiller_status
+ */
+@SubResource(parent = OrderResource2_2.class, path = "fulfillerdetails", supportedClass = FulfillerDetails.class, supportedOpenmrsVersions = {
+        "2.2.*" })
+public class FulfillerDetailsResource2_2 extends DelegatingSubResource<FulfillerDetails, Order, OrderResource2_2> {
+	
+	@Override
+	public FulfillerDetails newDelegate() {
+		return new FulfillerDetails();
+	}
+	
+	@Override
+	public FulfillerDetails save(FulfillerDetails delegate) {
+		Context.getOrderService().updateOrderFulfillerStatus(delegate.getOrder(), delegate.getFulfillerStatus(),
+		    delegate.getFulfillerComment());
+		return null;
+	}
+	
+	@Override
+	public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+		DelegatingResourceDescription delegatingResourceDescription = new DelegatingResourceDescription();
+		delegatingResourceDescription.addProperty("fulfillerStatus");
+		delegatingResourceDescription.addProperty("fulfillerComment");
+		return delegatingResourceDescription;
+	}
+	
+	@Override
+	public Model getCREATEModel(Representation rep) {
+		return new ModelImpl()
+		        .property("fulfillerComment", new StringProperty())
+		        .property("fulfillerStatus", new EnumProperty(Order.FulfillerStatus.class));
+	}
+	
+	@Override
+	public Order getParent(FulfillerDetails instance) {
+		return instance.getOrder();
+	}
+	
+	@Override
+	public void setParent(FulfillerDetails instance, Order parent) {
+		instance.setOrder(parent);
+	}
+	
+	@Override
+	public PageableResult doGetAll(Order parent, RequestContext context) throws ResponseException {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	public FulfillerDetails getByUniqueId(String uniqueId) {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	protected void delete(FulfillerDetails delegate, String reason, RequestContext context) throws ResponseException {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	public void purge(FulfillerDetails delegate, RequestContext context) throws ResponseException {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	public DelegatingResourceDescription getCreatableProperties() {
+		return getRepresentationDescription(null);
+	}
+}
--- a/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
+++ b/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
@@ -1,0 +1,38 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_2;
+
+import org.openmrs.Order;
+import org.openmrs.module.webservices.rest.web.RestConstants;
+import org.openmrs.module.webservices.rest.web.annotation.Resource;
+import org.openmrs.module.webservices.rest.web.representation.DefaultRepresentation;
+import org.openmrs.module.webservices.rest.web.representation.FullRepresentation;
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_10.OrderResource1_10;
+
+@Resource(name = RestConstants.VERSION_1 + "/order", supportedClass = Order.class, supportedOpenmrsVersions = { "2.2.*" })
+public class OrderResource2_2 extends OrderResource1_10 {
+	
+	/**
+	 * @see org.openmrs.module.webservices.rest.web.resource.impl.BaseDelegatingResource#getRepresentationDescription(org.openmrs.module.webservices.rest.web.representation.Representation)
+	 */
+	@Override
+	public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+		DelegatingResourceDescription delegatingResourceDescription = super.getRepresentationDescription(rep);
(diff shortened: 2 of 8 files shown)


fruether:
Actually the [FulfillerDetailsResource2_2Test](https://github.com/openmrs/openmrs-module-webservices.rest/blob/6a1374ccbd1ac1f6338537f9ca9fd5596d41be65/omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2Test.java) is now not working.

I am not sure if it is possible to test this subresource in this way? After all it does not have an own Uuid...

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
```diff
@@ -0,0 +1,107 @@
...
+	
+	@Override
+	public Model getCREATEModel(Representation rep) {
+		return new ModelImpl()
+				.property("fulfillerComment", new StringProperty())
+				.property("fulfillerStatus", new EnumProperty(null));
+	}
+	
+	@Override
+	public Model getGETModel(Representation rep) {
```
wluyima: I don't think this resource should support GET

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
```diff
@@ -0,0 +1,107 @@
...
+	
+	@Override
+	public FulfillerDetails getByUniqueId(String uniqueId) {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	protected void delete(FulfillerDetails delegate, String reason, RequestContext context)
+			throws ResponseException {
+		Context.getOrderService().updateOrderFulfillerStatus(delegate.getOrder(), null, null);
```
wluyima: I think this resource should not support any other operation exception save i.e. POST, if the client wishes to clear the fields, they can posting to the resource with blank values, the result would be the same.

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2WebTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.openmrs.module.webservices.rest.SimpleObject;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
+import org.openmrs.module.webservices.rest.web.v1_0.RestTestConstants2_2;
+import org.openmrs.module.webservices.rest.web.v1_0.controller.MainResourceControllerTest;
+import org.springframework.mock.web.MockHttpServletRequest;
+
+
+public class FulfillerDetailsResource2_2WebTest extends MainResourceControllerTest {
+	
+	@Test
```
wluyima: I don't think you need a new test method

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2WebTest.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+	
+	@Override
+	public String getUuid() {
+		return null;
+	}
+	
+	@Override
+	public long getAllCount() {
+		return 1;
```
wluyima: I believe this is not supported

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
```diff
@@ -0,0 +1,107 @@
...
+	}
+	
+	@Override
+	public void purge(FulfillerDetails delegate, RequestContext context) throws ResponseException {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	public DelegatingResourceDescription getCreatableProperties() {
+		DelegatingResourceDescription d = new DelegatingResourceDescription();
```
wluyima: calling return getRepresentationDescription(null) here actually avoid duplication since the properties are the same

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2WebTest.java
fruether: I think so because all other default test methods return a _ResourceDoesNotSupportOperationException_ exception

wluyima: You're right, I was thinking this is the test for the Resource

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
fruether: well I left it as it is because I changed getRepresentationDescription to actually add properties that are necessary to pass the test cases

fruether:
@wluyima I applied all your review comments. However, for _FulfillerDetailsResource2_2Test_ to pass I was forced to add uuid, selfLink and full properties to the object. I am not sure if this was a good idea?

If I do not not add these it looks like a test with the help of _BaseDelegatingResourceTest_ would not be possible since the super class checks for this properties. So I have the options of:
(1) Not testing it
(2) Adding these properties.

I choose (2) so far. What would you recommend?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
```diff
@@ -0,0 +1,104 @@
...
+		    delegate.getFulfillerComment());
+		return delegate;
+	}
+	
+	@Override
+	public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+		DelegatingResourceDescription delegatingResourceDescription = new DelegatingResourceDescription();
+		delegatingResourceDescription.addProperty("fulfillerStatus");
+		delegatingResourceDescription.addProperty("fulfillerComment");
+		delegatingResourceDescription.addProperty("uuid");
```
wluyima: You don't need these properties because they actually don't exist nor apply

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
```diff
@@ -0,0 +1,104 @@
...
+	}
+	
+	@Override
+	public FulfillerDetails getByUniqueId(String uniqueId) {
+		throw new ResourceDoesNotSupportOperationException();
+	}
+	
+	@Override
+	protected void delete(FulfillerDetails delegate, String reason, RequestContext context) throws ResponseException {
+
```
wluyima: You should throw ResourceDoesNotSupportOperationException here

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
```diff
@@ -0,0 +1,38 @@
...
+import org.openmrs.Order;
+import org.openmrs.module.webservices.rest.web.RestConstants;
+import org.openmrs.module.webservices.rest.web.annotation.Resource;
+import org.openmrs.module.webservices.rest.web.representation.DefaultRepresentation;
+import org.openmrs.module.webservices.rest.web.representation.FullRepresentation;
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_10.OrderResource1_10;
+
+@Resource(name = RestConstants.VERSION_1 + "/order", supportedClass = Order.class, supportedOpenmrsVersions = { "2.2.*" }, order = 1)
```
wluyima: Does the order attribute add anything here? I doubt

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsController2_2Test.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+	
+	@Override
+	public String getUuid() {
+		return null;
+	}
+	
+	@Override
+	public long getAllCount() {
+		return 1;
```
wluyima: I believe this should be zero

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
fruether: Then I am not able to create the testcases for  FulfillerDetailsRessource2_2 because they would assume these variables. If you yet think that they are not needed I will remove them

wluyima: What happens when you exclude them?

fruether: Then the tests in FulfillerDetailsResource2_2Test are failing e.g. https://github.com/openmrs/openmrs-module-webservices.rest/blob/e5e4c7d67cb0026d7ea99514fbad3e602f8a1f68/omod-common/src/test/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingResourceTest.java#L123

wluyima: That's what I was guessing, the correct thing to do in such cases is to override the tests and make sure they test for the correct thing i.e. that those fields do not exist since they are not include in the resource rep description.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/FulfillerDetailsResource2_2.java
wluyima: Those props should not be added

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
wluyima: Strange, there must be something not done correctly, I know various resources with the same name and no order attribute and they work e.g `LocationResource1_8`, `LocationResource1_9` and `LocationResource2_0`

Changing to v2 IS NOT an option, as I mentioned earlier, it stands for the entire REST API version and that isn't about to change anytime soon.

wluyima: @djazayeri do you know how the order attribute is used on a resource?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
```diff
@@ -0,0 +1,38 @@
...
+import org.openmrs.Order;
+import org.openmrs.module.webservices.rest.web.RestConstants;
+import org.openmrs.module.webservices.rest.web.annotation.Resource;
+import org.openmrs.module.webservices.rest.web.representation.DefaultRepresentation;
+import org.openmrs.module.webservices.rest.web.representation.FullRepresentation;
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_10.OrderResource1_10;
+
+@Resource(name = RestConstants.VERSION_1 + "/order", supportedClass = Order.class, supportedOpenmrsVersions = { "2.2.*" }, order=2)
```
fruether: Situation: I pointed out that without order=2 as argument the tests are failing. 

@wluyima replied with: "

> Strange, there must be something not done correctly, I know various resources with the same name and no order attribute and they work e.g LocationResource1_8, LocationResource1_9 and LocationResource2_0
> 
> Changing to v2 IS NOT an option, as I mentioned earlier, it stands for the entire REST API version and that isn't about to change anytime soon."

@djazayeri do you know how the order attribute is used on a resource?

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
wluyima: @fruether my guess is that you have both OrderResource_1_10 and OrderResource_2_2 both supporting OpenMRS version 2.2.x, I think you need to drop 2.2.x version from OrderResource_1_10 as a supported version and that should fix the issue without the order attribute. And if you think about, OrderResource_1_10 can't support 2.2.x since the new fullfiller details don't exist on it.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/OrderResource2_2.java
wluyima: @fruether there is a misconfiguration that I commented on that is causing the issue, you don't need the order attribute

fruether:
It worked now @wluyima I think all the comments are applied now 🤔

wluyima:
You need to pass the enum class to the EnumProperty

fruether:
@wluyima I guess all the other comments are applied now

PR #348, opened 2018-09-06
RESTWS-718: Fetch orders by orderType for patient

dsurrao:
https://issues.openmrs.org/browse/RESTWS-718

### DIFF
--- a/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
+++ b/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
@@ -23,6 +23,7 @@ import io.swagger.models.properties.StringProperty;
 import org.apache.commons.lang.StringUtils;
 import org.openmrs.CareSetting;
 import org.openmrs.Order;
+import org.openmrs.OrderType;
 import org.openmrs.Patient;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.webservices.docs.swagger.core.property.EnumProperty;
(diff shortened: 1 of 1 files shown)


mogoodrich:
Looks like there are some existing tests here... can you add some to include orderType:
https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/OrderController1_10Test.java

dsurrao:
OrderController1_10Test.java doesn't really test the REST endpoint, does it? Looks like I'll have to write tests that call methods of OrderResource1_10, and put those tests in OrderResource1_10Test.java

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
```diff
@@ -232,20 +233,33 @@ protected PageableResult doSearch(RequestContext context) {
...
 			CareSetting careSetting = null;
 			Date asOfDate = null;
+			OrderType orderType = null;
 			if (StringUtils.isNotBlank(asOfDateString)) {
 				asOfDate = (Date) ConversionUtil.convert(asOfDateString, Date.class);
 			}
 			if (StringUtils.isNotBlank(careSettingUuid)) {
 				careSetting = ((CareSettingResource1_10) Context.getService(RestService.class).getResourceBySupportedClass(
 				    CareSetting.class)).getByUniqueId(careSettingUuid);
+				if (careSetting == null) {
```
wluyima: Why are you adding this?

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
```diff
@@ -232,20 +233,33 @@ protected PageableResult doSearch(RequestContext context) {
...
+					return new EmptySearchResult();
+				}
+			}
+			if (StringUtils.isNotBlank(orderTypeUuid)) {
+				orderType = ((OrderTypeResource1_10) Context.getService(RestService.class).getResourceBySupportedClass(
+				    OrderType.class)).getByUniqueId(orderTypeUuid);
+				if (orderType == null) {
```
wluyima: Why do you return empty result?

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
dsurrao: The intention is that if no care setting uuid is passed in, return results matching all care settings, but if an invalid care setting is passed in, do not return any results.

The current functionality returns orders in all care settings if an invalid care setting is passed in, which seems incorrect.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
dsurrao: Similar to the case with care settings, if an invalid order type is passed in, it should return an empty result rather than return all orders.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
wluyima: If an invalid value is passed in, it's misleading to a client to send back an empty list, you should communicate back to them the problem by failing out loud.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java
wluyima: See my comment to that, it's misleading to send back an empty list.

mogoodrich: @wluyima I had been working on this with @dsurrao, and we went back and forth on this a bit.  We agreed what throwing an error seemed like the best solution, but noticed there was precedence for just returning an an empty set earlier in the list:

https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/OrderResource1_10.java#L218

So we just went with that for consistent, but are fine with changing them to fail loudly... it looking at another file, it looks like throwing an IllegalArgumentExcepotion is the current thing to do?

wluyima: That existing code that returns an empty list is doing the wrong thing too.

wluyima: I think it's better to throw `ObjectNotFoundException`

dsurrao: Thanks for the feedback! I've created a new PR with these updates.

PR #338, opened 2018-08-28
RESTWS-712: Posting a fullRepresentation accepted

kaweesi:

PR #350, opened 2018-09-11
RESTWS-712: Posting a fullRepresentation accepted

kaweesi:
- originally at: https://github.com/openmrs/openmrs-module-webservices.rest/pull/338


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
```diff
@@ -366,7 +367,8 @@ public static Object getGroupMembers(Obs obs) throws ConversionException {
 	 */
 	@PropertySetter("concept")
 	public static void setConcept(Obs obs, Object value) {
-		obs.setConcept(Context.getConceptService().getConceptByUuid((String) value));
+		String uuid = value instanceof Map ? ((Map) value).get("uuid").toString() : (String) value;
+		obs.setConcept(Context.getConceptService().getConceptByUuid(uuid));
```
wluyima: In theory you should always delegate resource lookup by uuid to the resource

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientResource1_8.java
```diff
@@ -225,6 +226,7 @@ public Object create(SimpleObject propertiesToCreate, RequestContext context) th
 	}
 	
 	public Patient getPatient(SimpleObject propertiesToCreate) {
+		removeKeys(propertiesToCreate, "preferredName", "preferredAddress", "deathdateEstimated", "birthtime");
```
wluyima: The framework should be ignoring these, if it doesn't we need to fix it instead of doing it here

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
```diff
@@ -31,7 +33,8 @@
 	 */
 	@PropertySetter("concept")
 	public static void setConcept(Obs obs, Object value) {
-		obs.setConcept(new ConceptResource1_9().getByUniqueId((String) value));
+		String uuid = value instanceof Map ? ((Map) value).get("uuid").toString() : (String) value;
+		obs.setConcept(new ConceptResource1_9().getByUniqueId(uuid));
```
wluyima: Same here, you should be delegating to the resource

wluyima: Actually, this is close to what I was suggesting, you only need to change it to look up the resource rather than instantiating one from here

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ProblemResource1_9.java
```diff
@@ -32,7 +34,8 @@
 	
 	@PropertySetter("problem")
 	public static void setProblem(Problem problem, Object value) {
-		problem.setProblem(new ConceptResource1_9().getByUniqueId((String) value));
+		String uuid = value instanceof Map ? ((Map) value).get("uuid").toString() : (String) value;
```
wluyima: Possibly you can add a utility function to do this since you are repeating it several times

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
```diff
@@ -296,11 +297,37 @@ public static Object convert(Object object, Type toType) throws ConversionExcept
...
+		Converter converter = null;
 		Object ret = null;
 		Object uuid = map.get(RestConstants.PROPERTY_UUID);
+		
+		if (toClass.equals(String.class)) {//is referenced object, retrieve by URI
+			String uri = getSelfLink(map);
+			//TODO RestConstants.URI_PREFIX may be poiting to another server?
+			String resourceCode = uri.split("/ws/rest/v1/")[1].replace("/" + uuid.toString(), "");
+			OpenmrsObject o = null;
+			if ("concept".equals(resourceCode)) {
```
wluyima: I think this needs to be a more reliable in order to work for all resources.

I would also think that the correct thing that needs to happen here is to fetch this resource from the remote server because the referenced instance won't always exist on the machine where this is getting executed.

wluyima: I'm not sure if we want to do this though, I suggest we leave this one out in the meantime because it's very likely to be tricky to implement.

omod-1.8/src/test/resources/create_patient_from_get.json
```diff
@@ -0,0 +1,267 @@
...
+            "resourceVersion":"1.8"
+        }
+    ],
+    "person":{
+        "uuid":"a65fc361-f6ac-40ba-8485-94749c061509",
+        "display":"Jonnea Bijo",
+        "gender":"M",
+        "age":23,
+        "birthdate":"1995-01-01T00:00:00.000+0000",
+        "birthdateEstimated":true,
```
wluyima: This was added in 1.11 so it should not be here, it's one of the reasons why the test fails since it is a 1.9 test

omod-1.8/src/test/resources/create_patient_from_get.json
```diff
@@ -0,0 +1,267 @@
...
+                        "rel":"self",
+                        "uri":"http://192.168.1.100/openmrs/ws/rest/v1/user/1c3db49d-440a-11e6-a65c-00e04c680037"
+                    }
+                ]
+            },
+            "dateCreated":"2018-08-26T10:40:30.000+0000",
+            "changedBy":null,
+            "dateChanged":null
+        },
+        "deathdateEstimated":false,
```
wluyima: This was added in 1.11 so it should not be here, it's one of the reasons why the test fails since this file is in the 1.8 subproject and the test where it is used is a 1.9 test

omod-1.8/src/test/resources/create_patient_from_get.json
```diff
@@ -0,0 +1,267 @@
...
+        "birthtime":null,
```
wluyima: Same here, this was added in 1.10 so it should not be here, it's one of the reasons why the test fails since this file is in the 1.8 subproject and the test where it is used is a 1.9 test

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientResource1_8.java
wluyima: deathdateEstimated and birthtime were added like in 1.11 and 1.10 respectively, see my comments below, meaning you should not be worrying about them here since this is a 1.8 resource, it is just the test data that is incorrect

omod-1.9/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/CreateUpdatePatientResource1_9Test.java
```diff
@@ -73,4 +75,18 @@ public void shouldUpdatePatient() throws Exception {
 		Assert.assertEquals("Race = Muslim", attributes.get(0).get("display"));
 	}
 	
+	@Test
+	public void shouldCreatePatient_fromGET() throws Exception {
+		executeDataSet("personAttributeTypeWithConcept.xml");
+		SimpleObject patientSimpleObject = new SimpleObject();
+		InputStream object = getClass().getClassLoader().getResourceAsStream("create_patient_from_get.json");
```
wluyima: One of the reasons this test fails is because of the test patient data in this json file, this is a 1.9 test yet the test json patient data includes fields that were added in later versions hence the failure.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
wluyima: In fact, this isn't the job of the rest module, if a property value is a reference i.e. a resource URL, the client should resolve that before it submits the data, it means this logic belongs in the client which in your use case will be the SYNC2 module, makes sense?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingResource.java
```diff
@@ -118,14 +119,18 @@ public Model getGETModel(Representation rep) {
...
-		propertiesIgnoredWhenUpdating.add("display");
-		propertiesIgnoredWhenUpdating.add("links");
-		propertiesIgnoredWhenUpdating.add("auditInfo");
-		propertiesIgnoredWhenUpdating.add(RestConstants.PROPERTY_FOR_RESOURCE_VERSION);
+		propertiesIgnoredOnPostAndPut.add("display");
+		propertiesIgnoredOnPostAndPut.add("links");
+		propertiesIgnoredOnPostAndPut.add("auditInfo");
+		propertiesIgnoredOnPostAndPut.add(RestConstants.PROPERTY_FOR_RESOURCE_VERSION);
+		
+		for (int i = 7; i < 16; i++) {
```
wluyima: This is also unnecessary, it's the test data that is wrong, the higher addresses fields 7-15 were added in 2.0

omod-1.8/src/test/resources/create_location_from_get.json
```diff
@@ -0,0 +1,54 @@
...
+            "display":"admin"
+        },
+        "dateCreated":"2018-08-26T12:50:35.000+0000",
+        "changedBy":null,
+        "dateChanged":null
+    },
+    "attributes":[
+
+    ],
+    "address7":null,
```
wluyima: This was added in 2.0 so it should not be here, it's one of the reasons why the test fails since this file is in the 1.8 subproject and the test where it is used is a 1.9 test

omod-1.8/src/test/resources/create_patient_from_get.json
```diff
@@ -0,0 +1,267 @@
...
+            "address3":null,
+            "address4":null,
+            "address5":null,
+            "address6":null,
+            "startDate":null,
+            "endDate":null,
+            "latitude":null,
+            "longitude":null,
+            "voided":false,
+            "address7":null,
```
wluyima: Same here, these were added in 2.0 so it should not be here, it's one of the reasons why the test fails since this file is in the 1.8 subproject and the test where it is used is a 1.9 test

omod-1.8/src/test/resources/create_patient_from_get.json
```diff
@@ -0,0 +1,267 @@
...
+                "address3":null,
+                "address4":null,
+                "address5":null,
+                "address6":null,
+                "startDate":null,
+                "endDate":null,
+                "latitude":null,
+                "longitude":null,
+                "voided":false,
+                "address7":null,
```
wluyima: Same for here

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientResource1_8.java
wluyima: Removing of preferredName and preferredAddress is the responsibility of the client and not the rest API

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
wluyima: If you think about, what the code in this method is doing is already taken care of by the framework and now all is doing is causing bugs like the one you are running into, the fix is to remove it anyways.

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ProblemResource1_9.java
wluyima: See my comment above, this needs to just go away

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
wluyima: Think about, it's very possible that the referenced resource doesn't exist on this server even though it does exist on the one where it was fetched from, it means it has to be fetched from the source and not this instance like you are doing.

dkayiwa:
@kaweesi are you still interested in finishing this up.

dkayiwa:
Ping me to reopen this when you resume working on it.

PR #352, opened 2018-09-25
Updated MainResourceController to only purge a Resource if purge para…

samuelmale:
…m is equal to true

Ticket: https://issues.openmrs.org/browse/RESTWS-703

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainResourceController.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainResourceController.java
@@ -152,7 +152,7 @@ public class MainResourceController extends BaseRestController {
 	 * @param response
 	 * @throws Exception
 	 */
-	@RequestMapping(value = "/{resource}/{uuid}", method = RequestMethod.DELETE, params = "purge")
+	@RequestMapping(value = "/{resource}/{uuid}", method = RequestMethod.DELETE, params = "purge=true")
 	@ResponseBody
 	public Object purge(@PathVariable("resource") String resource, @PathVariable("uuid") String uuid,
 	        HttpServletRequest request, HttpServletResponse response) throws ResponseException {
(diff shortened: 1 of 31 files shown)


samuelmale:
Could you please look at this! I'm not sure why we have the above issues with the POM file yet I didn't even touch it.

@wluyima , @dkayiwa

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainResourceController.java
```diff
@@ -155,11 +154,19 @@ public Object delete(@PathVariable("resource") String resource, @PathVariable("u
 	@RequestMapping(value = "/{resource}/{uuid}", method = RequestMethod.DELETE, params = "purge")
```
wluyima: I believe changing the value of this **params** attribute is enough without touching the method code

wluyima: That way you don't have to bake the void/retire logic in the method.

samuelmale: I think I'm missing something @wluyima . 

>A resource should only be purged when the purge parameter is set to true

The above being the title of this ticket, how would this be possible without changing the method code accordingly. I think I'm not understanding you ..

wluyima: I think the params attribute can take a spring EL expression, I think in this case you need to set the value for the purge parameter to true, you might want to look at the spring docs for the params attribute on RequestMapping for the exact syntax

wluyima: That way spring will only invoke this controller if purge is set explicitly set to true by the client, you won't have to make any code changes in the controller method.

samuelmale:
Perfect, resolved all issues @wluyima

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainResourceController.java
```diff
@@ -152,7 +152,7 @@ public Object delete(@PathVariable("resource") String resource, @PathVariable("u
 	 * @param response
 	 * @throws Exception
 	 */
-	@RequestMapping(value = "/{resource}/{uuid}", method = RequestMethod.DELETE, params = "purge")
+	@RequestMapping(value = "/{resource}/{uuid}", method = RequestMethod.DELETE, params = "purge=true")
```
wluyima: There you go! See how you didn't have to make any changes within the controller method? And everything still works

PR #355, opened 2018-10-04
UHM-3497: Bug fix, do not return duplicate encounters for search by obs

dsurrao:
https://tickets.pih-emr.org/browse/UHM-3497

Fixed a bug where duplicate encounters were being returned from EncounterSearchHandler.


mogoodrich:
Merged... fyi @dsurrao generally for modules that have their own JIRA projects in OpenMRS JIRA, we should try to create tickets in those projects for the changes, and use a reference to that ticket, as opposed to one in PIH JIRA, in the commit.

PR #356, opened 2018-10-24
RESTWS-726: Return diagnoses with an encounter

dsurrao:
https://issues.openmrs.org/browse/RESTWS-726

### DIFF
--- a/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/EncounterResource2_2.java
+++ b/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/EncounterResource2_2.java
@@ -1,0 +1,44 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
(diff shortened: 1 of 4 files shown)


omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/EncounterResource2_2.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Resource(name = RestConstants.VERSION_1 + "/encounter", supportedClass = Encounter.class, supportedOpenmrsVersions = { "2.2.*" })
+public class EncounterResource2_2 extends EncounterResource1_9 {
+	
+	@Override
+	public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+		DelegatingResourceDescription description = null;
+		if (rep instanceof DefaultRepresentation || rep instanceof FullRepresentation) {
+			description = super.getRepresentationDescription(rep);
+			description.addProperty("diagnoses");
```
wluyima: Is it intentional not to include these among creatable/updatable properties?

mogoodrich:
Good question @wluyima ... this ticket was just about making sure that diagnoses were returned as part of an encounter when when requesting an encounter  since we noticed this was missing.  We didn't really give any thought to the create/update part, but I think it is likely we want to make diagnoses createable and updateable, but we can handle this under a separate ticket.

wluyima:
I think it should happen as part of the same ticket, it's how we have always handled the addition of new resources.
The issue is that the Sync2 module needs to sync encounters, it means we have to exclude diagnoses from an encounter when syncing which is not right. It is a trivial task anyways to add the property to creatable/updatable properties

PR #363, opened 2018-11-25
Filter based on content-type: only allow application/json

isears:


isears:
Discussion on security mailing list: https://talk.openmrs.org/c/software/security

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/AuthorizationFilter.java
```diff
@@ -70,6 +70,15 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 			HttpServletResponse httpresponse = (HttpServletResponse) response;
 			httpresponse.sendError(HttpServletResponse.SC_FORBIDDEN, "IP address '" + request.getRemoteAddr()
 			        + "' is not authorized");
+			return;
+		}
+		
+		// check content-type (enforce json-only api)
```
wluyima: - I don't think this belongs here, this is for authorization purposes, I'd put this in a separate filter.
- The API actually accepts xml too

PR #366, opened 2018-11-29
Update openmrsVersion to 2.2.0-alpha

samuelmale:
This fixes the other bug caused by my last commit @dkayiwa


dkayiwa:
@samuelmale this is not the right fix. Just go wherever openmrs.version.2.2.0-SNAPSHOT was used and change it to the expected

PR #369, opened 2018-12-01
RESTWS-742: Filter based on content-type

isears:


isears:
Discussion on security mailing list: https://talk.openmrs.org/c/software/security

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/ContentTypeFilter.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Filter intended for all /ws/rest calls that prevents non-json Content-Types due to security
```
dkayiwa: Instead of saying non-json content types, shouldn't we say xml content types? Because that is what we are disabling.

dkayiwa:
@isears i have tried to run these changes in the reference application. When i search and then select a patient, i get and error and this stack trace: https://pastebin.com/t5CFWW6b

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/ContentTypeFilter.java
isears: Good catch! I had initially started with the idea of blocking everything but json, but eventually settled on just blocking xml. I'll change this.

isears:
This is concerning. It's possible some parts of the refapp are using the api's XML functionality. I'll look into it.

isears:
Ok, I think I found the issue. It looks like the filter was dropping requests that had a null content-type (i.e. GET requests). I've added a fix and a test case. Thanks for checking that @dkayiwa

dkayiwa:
@isears did you try reproducing the reported security problem before and after these changes?

isears:
@dkayiwa yes: I'm verifying that XML payloads don't get deserialized by POST-ing `<map></map>` to `/openmrs/ws/rest/v1/obs` and checking the logs for the following error:

`org.springframework.beans.TypeMismatchException: Failed to convert value of type 'java.util.HashMap' to required type 'org.openmrs.module.webservices.rest.SimpleObject'`. 

The presence of a TypeMismatchException indicates that the system is attempting to deserialize XML payloads.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/AuthorizationFilter.java
```diff
@@ -70,6 +70,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
 			HttpServletResponse httpresponse = (HttpServletResponse) response;
 			httpresponse.sendError(HttpServletResponse.SC_FORBIDDEN, "IP address '" + request.getRemoteAddr()
 			        + "' is not authorized");
```
dkayiwa: Did you intentionally make changes to this class?

isears: Yes. I added the return statement because I found that deserialization was occurring even after an IP failed the "allowed IPs" check. I figured it makes sense to stop processing a request's payload after it fails the "allowed IPs" authorization check, but this y isn't strictly necessary to patch the vulnerability. If you think it's going to cause instability somewhere else in the system I can take it out.

dkayiwa: I have not deeply evaluated the effect of this. Just saw this comment on line 104 "continue with the filter chain in all circumstances". Does it matter if you just did chain.doFilter(request, response); before returning?

isears: It does, unfortunately: a lot of processing happens in the `chain.doFilter(request, response)` call tree, to include deserialization. The purpose of the return statement is to prevent all that and just go straight to sending an HTTP 403 (Forbidden) response when appropriate.

dkayiwa: Fair enough. 😊

isears:
@dkayiwa pushed the updates we talked about

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/filter/ContentTypeFilter.java
```diff
@@ -0,0 +1,78 @@
...
+	
+	/**
+	 * @see javax.servlet.Filter#doFilter(javax.servlet.ServletRequest,
+	 *      javax.servlet.ServletResponse, javax.servlet.FilterChain)
+	 */
+	@Override
+	public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException,
+	        ServletException {
+		
+		// check content-type (deny xml requests)
```
dkayiwa: For the above comment, wouldn't it be better to either remove it or say something like allow only json requests?

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/ContentTypeFilterTest.java
```diff
@@ -0,0 +1,82 @@
...
+	@Before
+	public void init() {
+		testFilter = new ContentTypeFilter();
+		mockChain = new MockFilterChain();
+		req = new MockHttpServletRequest();
+		resp = new MockHttpServletResponse();
+	}
+	
+	@Test
+	public void doFilter_shouldFilterXMLContent() throws IOException, ServletException {
```
dkayiwa: Since we have used "Allow" for the other tests, can we consistently use the same even for this test?
When i first looks at "Filter" in the name, it was not obvious to me whether it is to filter to allow or not to allow.

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/ContentTypeFilterTest.java
```diff
@@ -0,0 +1,82 @@
...
+		
+		List<String> xmlContentTypes = Arrays.asList("application/xml", "text/xml");
```
dkayiwa: Can we add at least one non xml content type to make it more explicit that we intentionally decided to disable not only xml, but other content types too, that are not json?

dkayiwa: With the above, even the test method name would change to reflect that we are disabling everything that is not json

isears:
@dkayiwa I agree with all your suggestions. Just pushed changes that I think should address everything.

omod-common/src/test/java/org/openmrs/module/webservices/rest/web/filter/ContentTypeFilterTest.java
```diff
@@ -0,0 +1,82 @@
...
+		req.setContentType("application/json");
+		req.setMethod("POST");
+		req.setRequestURI("/ws/rest/v1/obs");
+		testFilter.doFilter(req, resp, mockChain);
+		
+		Assert.assertNotEquals(resp.getStatus(), HttpServletResponse.SC_UNSUPPORTED_MEDIA_TYPE);
+	}
+	
+	@Test
+	public void doFilter_shouldAllowGetRequest() throws IOException, ServletException {
```
dkayiwa: As for the methods name, should allow Get request for all content types?

isears: According to the RFCs, a GET request shouldn't have a Content-Type, but I think an attacker could theoretically craft a GET request with a Content-Type header that may trigger this vulnerability. So I think two things are important here:

1. We make sure we're allowing RFC-compliant GET requests because they are crucial to the normal operation of the Refapp.
2. We make no assumptions about RFC-compliance when security checking and only allow requests with json or null Content-Types, regardless of the type of request (GET, POST, PUT, etc.)

With that in mind, would it be more appropriate to split that into two tests? E.g. doFilter_shouldAllowGetRequest and doFilter_shouldAllowNullContentType?

dkayiwa: makes sense

dkayiwa:
Thanks @isears for the ping and hard work! 👍

PR #374, opened 2019-01-04
 RESTWS-730 Added PersonResource1_10.java in omod-1.10 .

Andu033:
I added PersonResource in omod-1.10 to solve [RESTWS-730](https://issues.openmrs.org/browse/RESTWS-730). Also a test class. The Resource is just like the one in 1.11.


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ConceptResource1_8.java
```diff
@@ -54,17 +54,7 @@
...
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Locale;
-import java.util.Objects;
-import java.util.Set;
+import java.util.*;
```
mks-d: Why did you touch that file?

omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10Test.java
```diff
@@ -0,0 +1,47 @@
+//package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_10;
```
mks-d: This file is not properly indented.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,101 @@
...
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingCrudResource;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.response.ResourceDoesNotSupportOperationException;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PersonResource1_8;
+
+/**
+ * {@link Resource} for Person, supporting standard CRUD operations
+ */
+@Resource(name = RestConstants.VERSION_1 + "/person", order = 1, supportedClass = Person.class, supportedOpenmrsVersions = {
+        "1.10.*","1.11.*", "1.12.*", "2.0.*", "2.1.*" })
```
mks-d: Surely this is not correct since there is already a `PersonResource1_11`.
Shouldn't it just be?
```java
supportedOpenmrsVersions = { "1.10.*" }
```
Also not sure if the `order` is of any use since there shouldn't be any overlap.

@dkayiwa?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ConceptResource1_8.java
```diff
@@ -54,17 +54,7 @@
...
```
mks-d: Please make sure you don't touch that file, this is out of scope.

omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10Test.java
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_10;
```
mks-d: Please check the indentation on that file.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,101 @@
...
+        return description;
+    }
+
+    /**
+     * @see org.openmrs.module.webservices.rest.web.resource.impl.BaseDelegatingResource#getCreatableProperties()
+     */
+    @Override
+    public DelegatingResourceDescription getCreatableProperties() {
+        DelegatingResourceDescription description = super.getCreatableProperties();
+        description.addProperty("deathdateEstimated");
```
mks-d: I think that `deathdateEstimated` was only added in 1.11.x, it should not be known in the context of 1.10.x compatibility.

Andu033:
Ok. I will update that changes.

Andu033:
Some intent changes and also added  @resource with it's configuration.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,58 @@
...
+        return addNewProperties(super.getCREATEModel(rep), rep);
+    }
+
+    @Override
+    public Model getUPDATEModel(Representation rep) {
+        return addNewProperties(super.getUPDATEModel(rep), rep);
+    }
+
+    private Model addNewProperties(Model model, Representation rep) {
+
```
mks-d: Remove one blank line above.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,58 @@
...
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PersonResource1_8;
+
+/**
+ * {@link Resource} for Person, supporting standard CRUD operations
+ */
+@Resource(name = RestConstants.VERSION_1 + "/person", order = 1, supportedClass = Person.class, supportedOpenmrsVersions = {
+        "1.10.*"})
+public class PersonResource1_10 extends PersonResource1_8 {
+
+
```
mks-d: Remove one blank line above.

Andu033: Ok. I will work on that.

Andu033: @mks-d in 1.10.x there is no variable birthtime in Person.java . Should i add it?

dkayiwa:
@Andu033 did you address the changes requested by @mks-d ?

Andu033:
@dkayiwa I solved some of them, but he told me to add birth time property on resource and test it but in person 10.* is no birth time field. I checked in my ide the imported package. So I think that will solve the problem if I extend supported version of personresource1_8 to cover 10.*. What do you think?

dkayiwa:
Did you try it out?

Andu033:
@dkayiwa  yep. Because i tried to test getBirthtime() and i get there is no such method and i checked the Person.java class coresponding to this version and there was no Birthtime atribute and alsl no methods for them . By the way how i can see the differences between different versions of Person.java class?

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
mks-d: @Andu033 it's there: https://github.com/openmrs/openmrs-core/blob/1.10.x/api/src/main/java/org/openmrs/Person.java#L61

The link above is on branch **1.10.x**. Can you expand on what issue you're facing?

mks-d:
@Andu033 the getter for birth time on branch 1.10.x is here: https://github.com/openmrs/openmrs-core/blob/1.10.x/api/src/main/java/org/openmrs/Person.java#L253-L259

Again, can you expand on what issue you're experiencing with this member/method?

Andu033:
@mks-d yes. I investigated and my issue is that when I import org.openmrs.person; I get the one from branch 1_8 instead of 1_10. I don't know how to solve that problem.

mks-d:
@Andu033 this is because the subproject omod-1.10 depends on Core 1.10.2 (and not yet 1.10.3), see [here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/15300ebc372302200f74e592f0cb4bba7287cc22/omod-1.10/pom.xml#L14). Just up that dependency for `Person.getBirthtime()` to become known and available.

Remember also that in the end `PersonResource1_10` should only be loaded from Core 1.10.3, anything below that should still load `PersonResource1_8`.

Andu033:
@mks-d  i tested and birthtime is in 10.4 if i up to 10.3 birthtime is still not visibile . And another problem is that if i go up to 10.4 one test fail in drugresource test this test :getDrugsByMapping_shouldReturnADrugThatMatchesTheCodeAndSourceAndTheBestMapType
give me this result

Expected :05ec820a-d297-44e3-be6e-698531d9dd3f
Actual   :3cfcf118-931c-46f7-8ff6-7b876f0d4202

mks-d:
> i tested and birthtime is in 10.4 if i up to 10.3 birthtime is still not visibile

It's [here](https://github.com/openmrs/openmrs-core/blob/d0109e54f503d86323dac1a89eb4395d91ec0218/api/src/main/java/org/openmrs/Person.java#L268-L270), and that's off tag 1.10.3. Can you investigate further why depending on 1.10.3 does not bring that method signature?

Andu033:
> It's [here](https://github.com/openmrs/openmrs-core/blob/d0109e54f503d86323dac1a89eb4395d91ec0218/api/src/main/java/org/openmrs/Person.java#L268-L270), and that's off tag 1.10.3. Can you investigate further why depending on 1.10.3 does not bring that method signature?

I see , but i went in the maven repo , at openmrs 1.10.3 , decompiled the jar file and this is how Person.java looks like there https://pastebin.com/KNMPPpqk .
And if i get the code of imported file in my ide is the same as one in the link above. This could be a issue?

mks-d:
@Andu033 let's up the version to 1.10.4 then. Also you will see that I linked your ticket to [TRUNK-4727](https://issues.openmrs.org/browse/TRUNK-4727) that says that the addition of `getBirthtime()` was fixed as part of 1.10.4. If that's indeed true your ticket description will have to be amended.

I'm trying to find traces of the release 1.10.3 in our CI and couldn't find it, I wanted to make sure on which commit it had been based. Will keep looking tomorrow.

In the meantime, can you dig in and see why `DrugController1_10Test` is failing when you up the dependency to 1.10.4?

Andu033:
Ok. I will inspect the problem with drug controller.

Andu033:
I found what is the problem with this test . The test do a get request for a drug with some properties and get 2 drugs back , in 1.10.1-3 the drug we want to compare uuid is the second  but in 1.10.4 or higher the drug is first and we always compare with the second one so we get good response but in wrong order . I think a good fix for that would be to be more specific about properties and get just one drug, the one we want to get. Another fix is to sort the results but this is not consistent . What do you think?

Andu033:
Should I open a ticket for that?

Andu033:
Or to solve that problem I can make a set of drugs uuid and check if that one I want is in the set.

Andu033:
@mks-d i solved the problem with the test. My approach is not perfect but is much better than the previous one . Let me know if i should do more changes and tell me if you want me to come with better approach for the test i changed. Maybe we can merge  like this and if everything else is ok i will start a new issue with that test to see why we get drugs in other order.

omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/DrugController1_10Test.java
```diff
@@ -95,14 +96,20 @@ public void getDrugsByMapping_shouldReturnADrugThatMatchesTheCodeAndSourceAndThe
...
 		        "CD41003"), new Parameter("source", sourceUuid), new Parameter("preferredMapTypes", mapTypeUuids))));
-		
-		assertEquals(expectedDrugUuid, PropertyUtils.getProperty(Util.getResultsList(results).get(1), "uuid"));
+
+		// Checking if expected drug is contained in response of the request above
+		HashSet<Object> setOfDrugs=new HashSet<Object>();
+		for(Object drug:Util.getResultsList(results)) {
+			setOfDrugs.add(PropertyUtils.getProperty(drug,"uuid"));
+		}
+		assert setOfDrugs.contains(expectedDrugUuid);
```
mks-d: @Andu033 it seems your idea is correct, but can you confirm that whether pre or post-1.10.3 it is always the same set of drugs that is returned? Just ordered differently.

Unfortunately REST WS does not auto-format (yet), so you'll have to stick to the rules. See the third point in **8.2 Blank Spaces** [here](https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141388.html). With some renaming I would rather do:
```java
Set<Object> actualDrugs = new HashSet<Object>();
for(Object drug : Util.getResultsList(results)) {
  actualDrugs.add(PropertyUtils.getProperty(drug, "uuid"));
}
assert actualDrugs.contains(expectedDrugUuid);
```

Andu033: @mks-d  yes is the same set of drugs , i tested. The person who did the test wanted to test if 2 he get the drug from request but he saw that he got 2 drugs because 2 drugs matched these properties and said ok here is my drug the second one but he didn t checked for all versions .  And even if there was a different set of drugs each time my approach is still better because i check if the drug is returned by the request so if the drug exist in both places , he only check if the drug is the second one which is not what he wanted from the name of test (ShouldReturnADrugThatMaches....) a drug so he check if this return a drug that matches . He dont want to see if the drug is second he only want to see if the mapping return a drug . So my approach match 100% the test case. (if the drug is not returned by mapping fail, if it is pass and i also checked for other versions 1.10.3 post and pre).

Andu033:
I did the changes

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,81 @@
...
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingCrudResource;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PersonResource1_8;
+
+/**
+ * {@link Resource} for Person, supporting standard CRUD operations
+ */
+@Resource(name = RestConstants.VERSION_1 + "/person", order = 3, supportedClass = Person.class, supportedOpenmrsVersions = {
+        "1.10.4", "1.10.5", "1.10.6" })
```
mks-d: It would be nice if Core's `matchRequiredVersions` ([here](https://github.com/openmrs/openmrs-core/blob/4a0feb8da351088f25fdc4e6d324a1f277aa3410/api/src/main/java/org/openmrs/module/ModuleUtil.java#L292)) could do something like this:
```java
supportedOpenmrsVersions = {"1.10.4+"}
```
It doesn't but this would work:
```java
supportedOpenmrsVersions = {"1.10.4 - 1.10.6"}
```
I guess that's fine since there is little chances to have any further releases on 1.10.x.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonResource1_8.java
```diff
@@ -50,7 +50,7 @@
  * {@link Resource} for Person, supporting standard CRUD operations
  */
 @Resource(name = RestConstants.VERSION_1 + "/person", order = 2, supportedClass = Person.class, supportedOpenmrsVersions = {
-        "1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*", "2.2.*" })
+        "1.8.*", "1.9.*", "1.10.0", "1.10.1", "1.10.2", "1.10.3" })
```
mks-d: Same remark, use:
```java
supportedOpenmrsVersions = {"1.8.*", "1.9.*", "1.10.0 - 1.10.3"}
```

Andu033:
@mks-d  i did the changes

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
Andu033: i did

Andu033:
@mks-d  or @dkayiwa  can you check if everything is ok and merge it ?

omod-1.10/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_10/DrugController1_10Test.java
```diff
@@ -95,14 +97,20 @@ public void getDrugsByMapping_shouldReturnADrugThatMatchesTheCodeAndSourceAndThe
...
+
 		assertEquals(0, Util.getResultsSize(results));
 		
 		mapTypeUuids = conceptService.getConceptMapType(2).getUuid();
 		results = deserialize(handle(newGetRequest(getURI(), new Parameter("s", "getDrugsByMapping"), new Parameter("code",
 		        "CD41003"), new Parameter("source", sourceUuid), new Parameter("preferredMapTypes", mapTypeUuids))));
-		
-		assertEquals(expectedDrugUuid, PropertyUtils.getProperty(Util.getResultsList(results).get(1), "uuid"));
+
+		// Checking if expected drug is contained in response of the request above
```
mks-d: You can remove that comment, the code is well written and self-explanatory.

Andu033: You are right is pretty clear . I deleted it . Thanks :)

Andu033:
@mks-d is now OK?

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,81 @@
...
+public class PersonResource1_10 extends PersonResource1_8 {
+
+    /**
+     * @see DelegatingCrudResource#getRepresentationDescription(Representation)
+     */
+    @Override
+    public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+        DelegatingResourceDescription description = super.getRepresentationDescription(rep);
+        if (description != null) {
+            description.addProperty("birthtime");
```
mks-d: @Andu033 it just occurred to me that `PersonResource1_11` should probably now extend `PersonResource1_10`, and thereby stop adding the property `"birthtime"`, see [here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/527db5acb23147a6e4fc6001cb4664037b116a1c/omod-1.11/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_11/PersonResource1_11.java#L37-L68).

Andu033:
Ok @mks-d  . Thanks for help. I reverted the files with the tabs removed . Is anything else i should change?

mks-d:
Thanks for your commitment (and patience) @Andu033!

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
```diff
@@ -0,0 +1,81 @@
...
+import org.openmrs.module.webservices.rest.web.representation.Representation;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingCrudResource;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
+import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.PersonResource1_8;
+
+/**
+ * {@link Resource} for Person, supporting standard CRUD operations
+ */
+@Resource(name = RestConstants.VERSION_1 + "/person", order = 3, supportedClass = Person.class, supportedOpenmrsVersions = {
+        "1.10.4 - 1.10.6" })
```
wluyima: I'd think this should be 1.10.3

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PersonResource1_8.java
```diff
@@ -50,7 +50,7 @@
  * {@link Resource} for Person, supporting standard CRUD operations
  */
 @Resource(name = RestConstants.VERSION_1 + "/person", order = 2, supportedClass = Person.class, supportedOpenmrsVersions = {
-        "1.8.*", "1.9.*", "1.10.*", "1.11.*", "1.12.*", "2.0.*", "2.1.*", "2.2.*" })
+	"1.8.*", "1.9.*", "1.10.0 - 1.10.3" })
```
wluyima: And I'd think this should be 1.10.0-1.10.2

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
mks-d: @wluyima true, the confusion came from the fact that there is also a _branch_ 1.10.3 (??), and on that branch `birthtime` is not part of `Person` and @Andu033 was doing locally `git checkout 1.10.3`, this took him to the branch and not the release tag, hence the erroneous conclusion as to when `birthtime` got added.

@wluyima @dkayiwa can this branch be removed if it has not use?

@Andu033, sorry, would you mind submit another PR based on @wluyima's changes?

Andu033: @mks-d yes . But we must remove the branch first to make the birthtime exist in 1.10.3?

mks-d: The branch in itself has no impact, what matters is to figure out for sure whether `Person` in the released Core 1.10.3 (so the WAR coming out of the Maven repo) did contain `birthtime` or not.

Didn't you say that you even unpacked the WAR to check that out?

@dkayiwa / @wluyima is it possible that, somehow, Core 1.10.3 was released off this branch 1.10.3 and not off the _tag_ 1.10.3? There's something dodgy here.

dkayiwa: I have just taken a look at noticed that the 1.10.3 release was not done correctly. So a 1.10.3 branch and a 1.10.3 tag were created instead of just the 1.10.3 tag.

mks-d: >I have just taken a look at noticed that the 1.10.3 release was not done correctly.

@dkayiwa but do you know if it's the tag or the branch that represents what was actually packaged as a WAR on the Maven repo? I think @Andu033 did the test that when you (Maven) depend on 1.10.3 `Person.birthtime` isn't there to be found...

wluyima: Tags are what matter, in any case by convention true maintenance branches are named with a **.x** prefix e.g. 1.10.x, 1.11.x etc. therefore I think 1.10.3 must be a stray branch that was created by mistake.

wluyima: @mks-d even if no war file was created for the 1.10.3 tag which I strongly doubt, we can still create it at anytime since it exists, this is one of the primary reasons why tags are created in the first place, to be able to recreate archives at that particular revision in case need arises.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
wluyima: Birthtime was also backported to 1.11 too, I think in 1.11.4, therefore PersonResource1_11 can't only just extend PersonResource1_10, it would also need to have some smarter logic to exclude the property by checking the openmrs version one is running.

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
dkayiwa: To avoid the confusion, i have just deleted the erroneously created 1.10.3 branch. So we now have only the 1.10.3 tag which does not have the Person.birthtime field: https://github.com/openmrs/openmrs-core/blob/1.10.3/api/src/main/java/org/openmrs/Person.java

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
mks-d: Correct, `birthtime` appears from 1.11.4...

@wluyima are we going to have a `PersonResource1_11` and `PersonResource1_11_4`??

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
Andu033: Thanks @dkayiwa  . This means that no more changes are required , right?

mks-d: @wluyima, @Andu033 will confirm but I think that the released WAR 1.10.3 does not contain `Person.birthtime`. If that's indeed the case we know that the tag 1.10.3 is not the commit off which the release was made. So yes we could re-release a WAR out of that tag, but which version would that be?

Shouldn't we assume that the Maven reality is such that 1.10.3 out there do not contain `Person.birthtime` while 1.10.4 do.

---

@Andu033 yes something must still be done in regards to the 1.11.3 vs 1.11.4 situation.

dkayiwa: According to this ticket https://issues.openmrs.org/browse/TRUNK-4727 the birthtime field starts with 1.10.4

Am i missing something to expect it in 1.10.3?

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
wluyima: You could add smart logic in PersonResource1_11 that includes the property depending on the version it's running on

omod-1.10/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_10/PersonResource1_10.java
wluyima: Looks like there was a mess around the 1.10.3 and yes it appears like the fix actually shows up in 1.10.4

wluyima: Really strange, from the git history the [backport](https://github.com/openmrs/openmrs-core/commit/cfc2f26a6e8042e8b9cb95b7fd53a44cf55f7f5c) happened between the 1.10.2 and 1.10.3 releases but for some reason the  fix doesn't show up in 1.10.3 tag. I'm suspicious about the 1.10.3 release.

wluyima: I now recall what happened, there was a security fix that was applied and we had special releases for the supported release lines which included 1.10.3, the releases were such that they only contained this fix and everything that had been backported to the respective maintenance branches got bumped to the next maintenance releases for each line. So yeah the commit was made earlier but was played forward due to these release hence this showing up in 1.10.4 and not 1.10.3.

PR #380, opened 2019-01-29
RESTWS-751: Add support of including POST path operations for Subresources

samuelmale:
Ticket : https://issues.openmrs.org/browse/RESTWS-751

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
@@ -62,6 +62,7 @@ import org.springframework.util.ReflectionUtils;
 
 import java.io.OutputStream;
 import java.io.PrintStream;
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -322,8 +323,11 @@ public class SwaggerSpecificationCreator {
 							method.invoke(resourceHandler, null, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID,
 							    new RequestContext());
 						}
-						catch (ResourceDoesNotSupportOperationException re) {
-							return false;
+						catch (InvocationTargetException e) {
+							if (e.getCause() instanceof ResourceDoesNotSupportOperationException) {
+								return false;
+							}
+							resourceHandler.save(null);	
 						}
 						catch (Exception ee) {
 							// if the resource doesn't immediate throw ResourceDoesNotSupportOperationException


omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -319,13 +319,13 @@ private boolean testOperationImplemented(OperationEnum operation, DelegatingReso
...
 							// to avoid saving data to the database, we pass a null SimpleObject
-							method.invoke(resourceHandler, null, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID,
+							method.invoke(resourceHandler, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID, null,
 							    new RequestContext());
 						}
-						catch (ResourceDoesNotSupportOperationException re) {
-							return false;
-						}
 						catch (Exception ee) {
+							if (ee.getCause() instanceof ResourceDoesNotSupportOperationException) {
```
dkayiwa: What is the exact class of exception when the cause is an instance of ResourceDoesNotSupportOperationException?

samuelmale: InvocationTargetException

> its caused if the underlying method throws an exception.

dkayiwa: Can you put that as a comment in the code? That is because just looking at the same class caught a few lines above, one wonders why you would need to do this.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -319,13 +319,15 @@ private boolean testOperationImplemented(OperationEnum operation, DelegatingReso
...
-							method.invoke(resourceHandler, null, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID,
+							method.invoke(resourceHandler, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID, null,
 							    new RequestContext());
 						}
-						catch (ResourceDoesNotSupportOperationException re) {
-							return false;
-						}
 						catch (Exception ee) {
+							// All exceptions thrown in the underlying method invoked; are wrapped in the InvocationTargetException
+							// Check whether the operation isn't supported
```
samuelmale: Is that satisfying?

samuelmale:
> Why don't you have another catch handler for InvocationTargetException?
> That way we do not risk the comment getting out of sync with the code. In other words, we can get rid of it.

Okie

samuelmale:
But I feel I don't consent with you here. Only because all potential exceptions for `method.invoke(args..)` are wrapped under `InvocationTargetException`. Meaning it's the only exception thrown here.

I suppose we remove this catch

`catch(Exception e) {..}`

and we replace it with 

`catch(InvocationTargetException e) {..}`

What your say?

Please refer : https://stackoverflow.com/questions/6020719/what-could-cause-java-lang-reflect-invocationtargetexception

dkayiwa:
What will happen for cases which are not InvocationTargetException?

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -319,11 +320,14 @@ private boolean testOperationImplemented(OperationEnum operation, DelegatingReso
 					} else {
 						try {
 							// to avoid saving data to the database, we pass a null SimpleObject
-							method.invoke(resourceHandler, null, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID,
+							method.invoke(resourceHandler, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID, null,
 							    new RequestContext());
 						}
-						catch (ResourceDoesNotSupportOperationException re) {
```
dkayiwa: How about for cases which throw ResourceDoesNotSupportOperationException?

samuelmale: Actually it's never thrown here. It gets wrapped in the InvocationTargetException in the underlying invoked method.

https://stackoverflow.com/questions/6020719/what-could-cause-java-lang-reflect-invocationtargetexception

dkayiwa: How sure are you that it never gets thrown? Looking at the fact the the original author probably had it thrown to put this catch clause. Most probably in some cases.

samuelmale: I'm so sure with this. 
When the `ResourceDoesNotSupportOperationException` is thrown in the invoked method eg. `#create(args..)` that causes the `InvocationTargetException`. 
So if we want the original Exception(cause), we invoke `e.getCause()`.

Please take time and look at the docs : https://docs.oracle.com/javase/7/docs/api/java/lang/reflect/InvocationTargetException.html

samuelmale:
Yes

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -319,11 +320,14 @@ private boolean testOperationImplemented(OperationEnum operation, DelegatingReso
 					} else {
 						try {
 							// to avoid saving data to the database, we pass a null SimpleObject
-							method.invoke(resourceHandler, null, RestConstants.SWAGGER_IMPOSSIBLE_UNIQUE_ID,
```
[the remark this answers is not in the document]
dkayiwa: How come that an IllegalArgumentException was not being thrown?

samuelmale:
How sure are you that it wasn't thrown? Are you trying to say the way I
have arranged the args wrong?

samuelmale:
I can revert, though I'm sure it's the right thing.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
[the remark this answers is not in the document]
dkayiwa: Are you able to reproduce it?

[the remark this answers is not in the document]
samuelmale: Yes.

Wat has been happening is, that Exception has always been thrown for all Subresources and handled here in [this catch](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java#L331) block. So only those that support the `save()` method would be added in Swagger doc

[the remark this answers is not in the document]
dkayiwa: What happens when you remove the Exception catch block?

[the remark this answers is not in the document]
samuelmale: Well I have test it out. Seems the order of arguments doesn't matter in this case cause the IllegalArgumentException was not thrown. So should I revert it to the original order of args?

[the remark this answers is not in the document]
dkayiwa: Yes unless you can produce the error.

PR #384, opened 2019-02-21
RESTWS-595: added a living patient search handler and added test units

mozzy11:
Ticket: https://issues.openmrs.org/browse/RESTWS-595

Added a search For only living patients using an optional parameter "includeDead"


dkayiwa:
Looks like you already have another pull request for the same.

PR #413, opened 2019-11-21
RESTWS-759, filter out DISCONTINUE orders

cioan:
@mogoodrich , here is the corresponding PR for filtering out DISCONTINUE orders. Thanks!


mogoodrich:
Looks good @cioan ... looks like some of the tests are failing in on Travis with compile errors, but I"m guessing that is because the OpenMRS 2.3.x changes we pushed up weren't in place when the tests first run.   Will merge and see what happens.

PR #418, opened 2019-12-10
RESTWS-763: Add current provider to the session

corneliouzbett:
## Description of what I changed

Added current user selected location and provider to the session

## Issue I worked on

see https://issues.openmrs.org/projects/RESTWS/issues/RESTWS-763

## Checklist: I completed these to help reviewers :)

- [x] My pull request only contains **ONE single commit**
(the number above, next to the 'Commits' tab is 1).

  No? -> [read here](https://wiki.openmrs.org/display/docs/Pull+Request+Tips) on how to squash multiple commits into one

- [x] My IDE is configured to follow the [**code style**](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.

  No? Unsure? -> [configure your IDE](https://wiki.openmrs.org/display/docs/How-To+Setup+And+Use+Your+IDE), format the code and add the changes with `git add . && git commit --amend`

- [x] I have **added tests** to cover my changes. (If you refactored
  existing code that was well tested you do not have to add tests)

  No? -> write tests and add them to this commit `git add . && git commit --amend`

- [x] I ran `mvn clean package` right before creating this pull request and
  added all formatting changes to my commit.

  No? -> execute above command

- [x] All new and existing **tests passed**.

  No? -> figure out why and add the fix to your commit. It is your responsibility to make sure your code works.

- [x] My pull request is **based on the latest changes** of the master branch.

  No? Unsure? -> execute command `git pull --rebase upstream master`

PR #421, opened 2020-01-28
RESTWS-758 Add location to session resource, with RESTful endpoints

brandones:

Adds sessionLocation to the user context object. The GET session endpoint now includes the location. The POST session endpoint now accepts a property `sessionLocation` which should be the UUID of a location. When POST is called, it also sets the location ID on the HttpSession, for compatibility with the appui endpoint.

This clashes with https://github.com/openmrs/openmrs-module-webservices.rest/pull/418 . Maintainers?

## Issue I worked on

https://issues.openmrs.org/browse/RESTWS-758

- [x] My pull request only contains **ONE single commit**
- [x] I have **added tests** to cover my changes. (If you refactored

### DIFF
--- a/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_9/SessionController1_9.java
+++ b/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_9/SessionController1_9.java
@@ -7,14 +7,10 @@
  * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
  * graphic logo is a trademark of OpenMRS Inc.
  */
-package org.openmrs.module.webservices.rest.web.v1_0.controller.openmrs1_8;
-
-import java.util.HashSet;
-import java.util.Locale;
-import java.util.Map;
-import java.util.Set;
+package org.openmrs.module.webservices.rest.web.v1_0.controller.openmrs1_9;
 
 import org.apache.commons.lang3.LocaleUtils;
+import org.openmrs.Location;
 import org.openmrs.api.APIException;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.webservices.rest.SimpleObject;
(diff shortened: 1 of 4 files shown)


omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_9/SessionController1_9.java
```diff
@@ -64,27 +67,42 @@ public Object get(WebRequest request) {
...
+				throw new APIException(" '" + localeStr + "' is not in the list of allowed locales.");
+			}
 		}
-		Set<Locale> allowedLocales = new HashSet<Locale>(Context.getAdministrationService().getAllowedLocales());
-		if (allowedLocales.contains(locale)) {
-			Context.setLocale(locale);
-		} else {
-			throw new APIException(" '" + localeStr + "' is not in the list of allowed locales.");
+		String locationUuid = body.get("sessionLocation");
+		if (locationUuid != null) {
```
mseaton: Assuming there are use cases where sessionLocation is optional, do we want to provide a mechanism to "unset" a previously selected sessionLocation?  I'm not 100% sure the proper way to handle this, but it would seem reasonable to check if "sessionLocation" is found as an attribute, but a null or empty value is set for it, then set the sessionLocation on the session to null.   Otherwise, I think this PR looks good to merge.

brandones: So the RESTful way to do it would be to have routes GET/POST/DELETE `/session/locale`, `session/location`, etc. But I think it would be fine just to let `"sessionLocation":""` mean "unset."

brandones: Ah, OpenMRS Core UserContext doesn't provide a way to unset location. See https://github.com/openmrs/openmrs-core/blob/6082fd6dc6c3c0c2b3ea78aab7143f0dff855215/api/src/main/java/org/openmrs/api/context/UserContext.java#L422

mseaton: Fair enough.  Seems like something we might want to fix in core, but out of scope for this ticket...

PR #424, opened 2020-03-09
RESTWS-767: SessionController does not support custom representations

mogoodrich:


mseaton:
@mogoodrich I made a few comments on the ticket.

PR #433, opened 2020-05-17
Refactor Task action resource to take in task definition uuid in addi…

gitcliff:
link:https://issues.openmrs.org/browse/RESTWS-764


ibacher:
@gitcliff Can you add a test showing that sending a POST to the `TaskActionController2_4` with the payload:

```
{ "action": "runtask", "tasks": [" + getTaskUuid() + "] }
``` 

runs the task? This should be pretty similar to the existing tests for task name.

tendomart:
@ibacher  is everything here valid up to this level ,if some other person is to pick on this task ?

PR #435, opened 2020-05-21
RESTWS-775:Expand order search handler to support searching on Access…

mogoodrich:
…ion Number and Order Number


mogoodrich:
Yeah not sure what's going on with the formatter... and I noticed the Travis stuff... I'll keep an eye on on OpenMRS CI and make sure the build passed there....

mogoodrich:
Ah, @mseaton  Travis is failing because the updated OpenMRS 2.3.1 is not available yet.

PR #477, opened 2021-01-07
RESTWS-816: ObsResource to not fail conversions of obs drugs and obs locations.

mogoodrich:
…Encounter and one or more the nested obs is of type "valueLocation" or "valueDrug"

### DIFF
--- a/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
+++ b/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
@@ -60,6 +60,7 @@ public class ObsResource1_9 extends ObsResource1_8 {
             return super.convertDelegateToRepresentation(delegate, rep);
         } catch (ConversionException e) {
             // hack to handle https://issues.openmrs.org/browse/RESTWS-816
+            // (if converting to a custom rep fails because an obs is valueDrug or valueLocation, return null for the value of that obs instead of failing hard)
             if (delegate.getValueDrug() != null || "org.openmrs.Location".equals(delegate.getComment())) {
                 rep.removeProperty("value");
                 SimpleObject result = super.convertDelegateToRepresentation(delegate, rep);
--- a/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
+++ b/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
@@ -9,15 +9,18 @@
  */
 package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_9;
 
-import java.util.Map;
-
 import org.openmrs.Concept;
 import org.openmrs.Obs;
+import org.openmrs.module.webservices.rest.SimpleObject;
 import org.openmrs.module.webservices.rest.web.ConversionUtil;
 import org.openmrs.module.webservices.rest.web.RestConstants;
 import org.openmrs.module.webservices.rest.web.annotation.PropertySetter;
 import org.openmrs.module.webservices.rest.web.annotation.Resource;
+import org.openmrs.module.webservices.rest.web.resource.impl.DelegatingResourceDescription;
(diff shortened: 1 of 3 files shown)


omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/ObsResource1_9.java
```diff
@@ -42,12 +45,31 @@ public static void setConcept(Obs obs, Object value) {
...
 		obs.setConcept(ConversionUtil.getConverter(Concept.class).getByUniqueId((String) identifier));
 	}
-	
+
+
+    @Override
+    protected SimpleObject convertDelegateToRepresentation(Obs delegate, DelegatingResourceDescription rep) {
+        try {
+            return super.convertDelegateToRepresentation(delegate, rep);
+        } catch (ConversionException e) {
```
mseaton, line 61: Is this the lowest down we can get?  Seems like there might be other cases where a conversion exception is thrown and these conditions are true (though relatively unlikely, is possible), and this would end up swallowing those.

mogoodrich: This was as low as I think I could get without hacking something into the shared superclass, which I didn't want to do.

Note that if the Obs does *not* have a value drug or a comment indicating it's a value location, it just re-throws the error.  And if it *is* one of those two, it just runs the conversion again removing the "value" property, without trapping an other error that may occur.  So it seemed rather unlikely.

I did also want to set it so it so it only trapped the error if it was a Custom Rep that was being requested, but that would need to have happened higher up and left me with less control with what I could actually return.

dkayiwa:
It would be great to have a controller test to reproduce the exact problem that you reported.

mogoodrich:
Thanks @dkayiwa ... I'll see if it's straightforward to create a controller test.

omod-1.9/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
```diff
@@ -0,0 +1,65 @@
...
+        // our hack fix (see https://issues.openmrs.org/browse/RESTWS-816) is to simply to return null as the value for value drug
+        SimpleObject result = (SimpleObject) ConversionUtil.convertToRepresentation(encounter, new CustomRepresentation("uuid,obs:(value:(names))"), new EncounterResource1_8());
+        List<SimpleObject> resultObsList = (List<SimpleObject>) result.get("obs");
+        assertThat(resultObsList.size(), is(2));
+        SimpleObject resultObs1 = resultObsList.get(0);
+        SimpleObject resultObs2 = resultObsList.get(1);
+        assertTrue(resultObs1.containsKey("value"));
+        assertTrue(resultObs2.containsKey("value"));
+        assertTrue(resultObs1.get("value") == null && resultObs2.get("value") != null
+                || resultObs1.get("value") != null && resultObs2.get("value") == null);
```
mks-d, line 62: It would be more readable to give each obs a UUID in the first place and identify which one is which on that basis to then assert that the drug one has a null value.

omod-1.9/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
```diff
@@ -0,0 +1,65 @@
...
+
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+public class ConversionUtil1_9Test extends BaseModuleWebContextSensitiveTest {
+
+
+    @Test
+    public void convertToRepresentation_shouldNotFailIfUnableToParseObsValueUsingCustomRep() {
```
mks-d, line 33: I would say what this really asserts, something like: `convertToRepresentation_shouldConvertObsDrugValueAsNull`.

mks-d (commented):
Thanks @mogoodrich, I have some minor comments, mainly around naming. I also renamed the PR headline to
>`ObsResource` to not fail conversions of obs drugs and obs locations.

mogoodrich:
@dkayiwa  @mks-d  made an updated commit based on your feedback, merging now, thank!

PR #553, opened 2022-05-02
Improving supportedResources to allow easily compatibility with new v…

kdaud:

@ibacher this is a follow up commit to what was done at https://github.com/openmrs/openmrs-module-webservices.rest/commit/7ec313cb7de970d86c616833af72c772948e8b44.


kdaud:
Have got notification from Bamboo indicating Platform 2.5.x is passing however,  1.10.x, 1.11.x, 1.12.x, 2.0.x are failing. Could this be merge have messed up with those respective branches? If yes we can revert the merge. cc: @ibacher

kdaud:
I tested the changes locally and they were passing all the branches but not sure though why breaking in Bamboo?

ibacher:
@kdaud Oh shoot! Yeah, this class is only intended to be loaded for one version that's why it wasn't previously updated

ibacher:
I've reverted this commit.

PR #573, opened 2023-03-30
RESTWS-909: Condition resource fails when using a custom representati…

mseaton:
…on of codedOrNonCoded

## Issue I worked on
see https://issues.openmrs.org/browse/RESTWS-909

### DIFF
--- a/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/converter/openmrs2_2/CodedOrFreeTextConverter.java
+++ b/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/converter/openmrs2_2/CodedOrFreeTextConverter.java
@@ -15,7 +15,9 @@ import org.openmrs.ConceptName;
 import org.openmrs.annotation.Handler;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.webservices.rest.SimpleObject;
+import org.openmrs.module.webservices.rest.web.ConversionUtil;
 import org.openmrs.module.webservices.rest.web.api.RestService;
+import org.openmrs.module.webservices.rest.web.representation.CustomRepresentation;
 import org.openmrs.module.webservices.rest.web.representation.DefaultRepresentation;
 import org.openmrs.module.webservices.rest.web.representation.FullRepresentation;
 import org.openmrs.module.webservices.rest.web.representation.Representation;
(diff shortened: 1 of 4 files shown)


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingResource.java
```diff
@@ -410,82 +410,14 @@ public SimpleObject asRepresentation(T delegate, Representation representation)
...
-			}
+			repDescription = ConversionUtil.getCustomRepresentationDescription((CustomRepresentation) representation);
+			return convertDelegateToRepresentation(delegate, repDescription);
 		}
 		
 		throw new ConversionException("Don't know how to get " + getClass().getSimpleName() + "(" + delegate.getClass()
 		        + ") as " + representation.getRepresentation(), null);
 	}
 	
-	/**
```
mseaton, line 423: This is moved to a public static method in ConversionUtil without any changes

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
```diff
@@ -504,5 +505,70 @@ public static SimpleObject getAuditInfo(Object delegate) {
 		
 		return ret;
 	}
-
```
mseaton, line 507: This is moved here to a public static method in ConversionUtil without any changes, previously was a private instance method on BaseDelegatingResource

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingResource.java
```diff
@@ -410,82 +410,14 @@ public SimpleObject asRepresentation(T delegate, Representation representation)
 		
 		// finally if it is a custom representation and not supported by any other handler
 		if (representation instanceof CustomRepresentation) {
-			repDescription = getCustomRepresentationDescription((CustomRepresentation) representation);
-			if (repDescription != null) {
-				return convertDelegateToRepresentation(delegate, repDescription);
-			}
+			repDescription = ConversionUtil.getCustomRepresentationDescription((CustomRepresentation) representation);
```
mseaton, line 413: My IDE was telling me to change this because repDescription will never be null

omod-2.2/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs2_2/ConditionController2_2Test.java
```diff
@@ -325,4 +327,30 @@ public void shouldReturnAllConditions() throws Exception {
...
 		Assert.assertEquals(3, conditions.size());
 	}
+
+	@Test
+	public void shouldGetACustomRepresentation() throws Exception {
+		MockHttpServletRequest request = request(RequestMethod.GET, getURI());
+		request.addParameter("patientUuid", "da7f524f-27ce-4bb2-86d6-6d1d05312bd5");
+		request.addParameter("includeInactive", "true");
+		request.addParameter("v", "custom:(uuid,display,clinicalStatus,onsetDate,endDate,additionalDetail,condition:(coded:(id,uuid)))");
+		SimpleObject result = deserialize(handle(request));
```
mseaton, line 337: Without the fix, this test fails here.  It throws an error trying to process this request.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/converter/openmrs2_2/CodedOrFreeTextConverter.java
```diff
@@ -30,25 +32,37 @@ public class CodedOrFreeTextConverter extends BaseDelegatingConverter<CodedOrFre
...
+				Object o = ConversionUtil.getPropertyWithRepresentation(instance, propertyName, property.getRep());
+				if (o != null) {
+					codedOrFreeText.add(propertyName, o);
+				}
+			}
 		}
-		
-		if (instance.getNonCoded() != null) {
-			codedOfFreeText.add("nonCoded", instance.getNonCoded());
+		else {
```
mseaton, line 47: I don't know why this is showing such an extensive change.  All of the pre-existing code was simply moved unchanged into this else block.  So the only thing I changed was to say - if the Representation is a CustomRepresentation, then process it differently.  If it not a custom representation, no behavior has changed here.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/converter/openmrs2_2/CodedOrFreeTextConverter.java
```diff
@@ -30,25 +32,37 @@ public class CodedOrFreeTextConverter extends BaseDelegatingConverter<CodedOrFre
...
-			        .getResourceBySupportedClass(ConceptName.class);
-			codedOfFreeText.add("specificName", conceptNameResource.asRepresentation(instance.getSpecificName(), rep));
-		}
-		
-		if (instance.getCoded() != null) {
-			ConceptResource2_0 conceptResource = (ConceptResource2_0) Context.getService(RestService.class)
-			        .getResourceBySupportedClass(Concept.class);
-			codedOfFreeText.add("coded", conceptResource.asRepresentation(instance.getCoded(), rep));
+		SimpleObject codedOrFreeText = new SimpleObject();
+		if (rep instanceof CustomRepresentation) {
```
mseaton, line 36: I don't know why the framework doesn't automatically handle this, but it doesn't seem to.  So here, I'm explicitly parsing the custom representation and using that to determine which properties of the CodedOrNonCoded to convert, and how.

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/converter/openmrs2_2/CodedOrFreeTextConverter.java
```diff
@@ -30,25 +32,37 @@ public class CodedOrFreeTextConverter extends BaseDelegatingConverter<CodedOrFre
...
+			CustomRepresentation customRep = (CustomRepresentation) rep;
+			DelegatingResourceDescription drd = ConversionUtil.getCustomRepresentationDescription(customRep);
+			for (String propertyName : drd.getProperties().keySet()) {
+				DelegatingResourceDescription.Property property = drd.getProperties().get(propertyName);
+				Object o = ConversionUtil.getPropertyWithRepresentation(instance, propertyName, property.getRep());
+				if (o != null) {
```
mseaton, line 42: Not sure if this is correct - should I include this even if it is null, since it is specified in the custom rep?

mogoodrich: Not sure, I would guess skipping in the correct behavior.  At first instinct it might made sense to do something like ' uuid:"" '  but I doubt we'd want to included a nested set of empty properties, so I think just skipping makes sense, unless someone else as an opinion otherwise.

mogoodrich (commented):
LGTM, and it appears this answers why I had never seen any problems like this with custom reps: this only applies to the CodedOrFreeText type, which is only currently used for Diagnoses and Conditions, which I've never tried to fetch via REST.

PR #599, opened 2024-03-01
RESTWS-931: Visit Search: add support to include parent locations

mogoodrich:
RESTWS-932: Visit Search: add "toStartDate" parameter

See:
https://openmrs.atlassian.net/browse/RESTWS-931
https://openmrs.atlassian.net/browse/RESTWS-932

## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-

  No? -> execute above command

### DIFF
--- a/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java
+++ b/omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java
@@ -325,14 +325,12 @@ public class VisitResource1_9 extends DataDelegatingCrudResource<Visit> {
 		String locationParameter = context.getRequest().getParameter("location");
 		String includeInactiveParameter = context.getRequest().getParameter("includeInactive");
 		String fromStartDate = context.getRequest().getParameter("fromStartDate");
-		String toStartDate = context.getRequest().getParameter("toStartDate");
 		String visitTypeParameter = context.getRequest().getParameter("visitType");
 		String includeParentLocations = context.getRequest().getParameter("includeParentLocations");
 		if (patientParameter != null || includeInactiveParameter != null || locationParameter != null
 		        || visitTypeParameter != null) {
 			Date minStartDate = fromStartDate != null ? (Date) ConversionUtil.convert(fromStartDate, Date.class) : null;
-			Date maxStartDate = toStartDate != null ? (Date) ConversionUtil.convert(toStartDate, Date.class) : null;
-			return getVisits(context, patientParameter, includeInactiveParameter, minStartDate, maxStartDate, locationParameter,
+			return getVisits(context, patientParameter, includeInactiveParameter, minStartDate, locationParameter,
 			    visitTypeParameter, includeParentLocations);
 		} else {
 			return super.search(context);
@@ -340,7 +338,7 @@ public class VisitResource1_9 extends DataDelegatingCrudResource<Visit> {
 	}
 	
 	private SimpleObject getVisits(RequestContext context, String patientParameter, String includeInactiveParameter,
-	        Date minStartDate, Date maxStartDate, String locationParameter, String visitTypeParameter, String includeParentLocations) {
+	        Date minStartDate, String locationParameter, String visitTypeParameter, String includeParentLocations) {
 		Collection<Patient> patients = patientParameter == null ? null : Arrays.asList(getPatient(patientParameter));
 		Collection<Location> locations = locationParameter == null ? null :
 				Boolean.parseBoolean(includeParentLocations) ?
@@ -351,7 +349,7 @@ public class VisitResource1_9 extends DataDelegatingCrudResource<Visit> {
 		boolean includeInactive = includeInactiveParameter == null ? true : Boolean.parseBoolean(includeInactiveParameter);
 		return new NeedsPaging<Visit>(Context.getVisitService().getVisits(visitTypes, patients, locations, null,
 		    minStartDate,
-		    maxStartDate, null, null, null, includeInactive, context.getIncludeAll()), context).toSimpleObject(this);
+		    null, null, null, null, includeInactive, context.getIncludeAll()), context).toSimpleObject(this);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java
```diff
@@ -323,29 +325,49 @@ public SimpleObject search(RequestContext context) throws ResponseException {
...
 		String fromStartDate = context.getRequest().getParameter("fromStartDate");
+		String toStartDate = context.getRequest().getParameter("toStartDate");
 		String visitTypeParameter = context.getRequest().getParameter("visitType");
+		String includeParentLocations = context.getRequest().getParameter("includeParentLocations");
 		if (patientParameter != null || includeInactiveParameter != null || locationParameter != null
 		        || visitTypeParameter != null) {
 			Date minStartDate = fromStartDate != null ? (Date) ConversionUtil.convert(fromStartDate, Date.class) : null;
-			return getVisits(context, patientParameter, includeInactiveParameter, minStartDate, locationParameter,
-			    visitTypeParameter);
+			Date maxStartDate = toStartDate != null ? (Date) ConversionUtil.convert(toStartDate, Date.class) : null;
```
mseaton, line 334: My only concern here is if this supports passing in a date like `2024-03-01` and the user is going to expect to get back visits inclusive of this date, I don't think this will happen.  I think it will only return visits that are on `2024-03-01 00:00:00` and earlier.

mseaton (commented):
LGTM, other than one comment to check up on.

omod-1.9/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java
mogoodrich: That is correct, @mseaton .  It's a fair point that this could be confusing, but I *suspect* that is consistent with how other date-based searches with the REST module work, but worth double-checking.

jecihjoy: An a consumer of date-based searches I usually add one day to the date value ie if I want data for `2024-03-01` and earlier I would pass `2024-03-02`. I don't know if we should be using such a hack or we should consider correcting date-based search logic.

mogoodrich: @mseaton  I'm going to stick with this as-is right now, because this does seem to be the general current behavior, instead of getting funky and converting "2022-01-01" to "2022-01-01T23:59:59"

That being said, I'll fix my bug in the appointment code PR around this.

@jecihjoy  the issue the the approach of using "2024-03-02" to mean "before" "2024-03-02" is that it would include encounters/viists that happened exactly on midnight on 2024-03-02... which rarely an issue in a real-time use case, but may be a problem with Retrospective Data Entry where all visits on a certain date without a timestamp get a visit date set to "2024-03-02T00:00:00".  Better to do "2024-03-01T23:59:59" to be safe.

mseaton: @mogoodrich I don't really think that the behavior is right though.  Where else is this done?  Maybe we should ticket separately as a broading thing to investigate and fix?

mogoodrich: @mseaton  that's the way the encounter search handler appears to work... or, more generally, since everything in OpenMRS/Java is a Date, when you pass in a string and parse it as a date you end up with "2023-02-02T00:00:00". And all the underlying API methods operate on DateTimes.   

We could ticket special casing  REST web services to parse a date without a time as a date-at-one-second-before-midnight, which I agree is what the consumer likely really wants. But I'm resistant to just bake this in in one place.  We'd also probably want to think about how time zones might affect what people really want?  Ie if the date comes in at 2023-02-02" assumedly that means they want one second before midnight in the server time zone?  We start getting into the whole time zone rats nest and probably want to handle all this as part of a global time zone plan.

mogoodrich: Of course your point it valid that requesting instant-before-midnight on the client side it definitely not what is wanted when server and client time zones don't match... :(

mogoodrich: Maybe I will add a flag around this...

mseaton: @mogoodrich , if timezones are important to a client application, then they already have what they need on the server-side, as long as they submit a date with a time and timezone component.  But I think it would be a broad bug in our webservices, if they accept a "to date" parameter as "yyyy-MM-dd", and this generally didn't include any results on that date unless they occurred exactly at midnight.  _That_ IMHO, is a bug.

mseaton: @ibacher / @dkayiwa ? ^^

mogoodrich: Yeah, I'm coming around to the idea of assuming "if the input string is yyyy-MM-dd" the start date should be datetime midnight in the servers timezone, and end date should be datetime last-second-of-day in the servers timezone.

This does fail though in the "servers are always in UTC" use case, which I believe is the Mekom standard.  But there are probably lots of places where that fails, and this is probably the best we can do.

mogoodrich:
Merging this in with *just* the additional for including parent locations (RESTWS-931)

PR #609, opened 2024-06-04
RESTWS-939: DiagnosisResource should support formFieldPath property

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-939

  No? -> execute above command


IamMujuziMoses:
@samuelmale  @mseaton @dkayiwa please review

mseaton (changes requested):
My first reaction is that I would you think you could accomplish adding this support without adding a new omod-2.5 submodule.  As far as I can see in the code, you never access this property directly, only via reflection (via the string property name).  You should be able to simply wrap those statements in a check as to whether version 2.5+ of core is running, rather than have an entire new submodule to support.

If we do find we need to keep the submodule, then I have feedback about changes that would be needed, but I would think we can eliminate it altogether and keep this far simpler with just a few conditional changes to the existing DiagnosisResource2_2

omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_2/DiagnosisResource2_2.java
```diff
@@ -98,6 +99,9 @@ public void purge(Diagnosis diagnosis, RequestContext requestContext) throws Res
 	public DelegatingResourceDescription getRepresentationDescription(Representation representation) {
 		if (representation instanceof DefaultRepresentation) {
 			DelegatingResourceDescription description = new DelegatingResourceDescription();
+			if (ModuleUtil.isOpenmrsVersionInVersions("2.5.* - 9.*")) {
+				description.addProperty("formNamespaceAndPath");
+			}
```
IamMujuziMoses, line 104: @mseaton were you suggesting changes something like this?

mseaton: Something like this, yes.  Obviously you'd need to test this resource against version < 2.5 and versions >= 2.5 to ensure it works.  It would be nice if there were other convenience methods in ModuleUtil that allowed you to just pass in the minimum version, but if this is the best available, that's fine.  I'd at least create a convenience method in this class and use that (eg. `supportsFormNamespaceAndPath() { return ModuleUtil.isOpenmrsVersionInVersions("2.5.* - 9.*") }`

IamMujuziMoses: Are there any convenience methods I could use to test this resource against versions >= 2.5?

samuelmale:
@IamMujuziMoses are you still working on this?

brandones:
@samuelmale Looks like @IamMujuziMoses asked for a re-review from @mseaton four days ago. However the PIH team is about to be gone all of next week. I think it would be helpful if you, @dkayiwa , @ibacher , or someone else who understands REST WS code could provide a review (including ensuring that the concerns Mike raised are addressed).

ibacher:
I'm actually going to take the opposite view @mseaton took above. I think we _should_ just create the 2.5 omod and go from there. Two reasons:

1. We either need to use reflection or some other mechanism like used here to test if the "formNamespaceAndPath" property is added, whereas with a 2.5 OMOD, we do the version check once (or once-per-context refresh).
2. The `FormRecordable` interface was added to more than just Diagnosis in 2.5. It was also added to: Order, Condition, Allergy, and PatientState so ideally, we'd cover all of those resources in a 2.5 OMOD, which I think gives us enough content to justify creating the new sub-module.

ibacher:
Code-wise, I think we're better-off exposing the `formFieldNamespace` and `formFieldPath` properties directly rather than the `formFieldAndNamespace` property for parity with the existing Obs resource, which also exposes them as two fields.

IamMujuziMoses:
@ibacher please review

omod-2.5/pom.xml
```diff
@@ -0,0 +1,215 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+		      <groupId>org.apache.tomcat</groupId>
+		      <artifactId>jasper</artifactId>
+		      <version>${apacheTomcatVersion}</version>
+		      <scope>provided</scope>
+	     </dependency>
+
```
ibacher, line 197: Could you ensure that the whitespace in this file is consistent? This uses both tabs and spaces.

omod-2.5/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_5/DiagnosisResource2_5.java
```diff
@@ -0,0 +1,87 @@
...
+public class DiagnosisResource2_5 extends DiagnosisResource2_2 {
+
+    /**
+     * @see DelegatingCrudResource#getRepresentationDescription(Representation)
+     */
+    @Override
+    public DelegatingResourceDescription getRepresentationDescription(Representation rep) {
+        DelegatingResourceDescription description = super.getRepresentationDescription(rep);
+        if (description != null) {
+            description.addProperty("formNamespaceAndPath");
```
ibacher, line 41: Remember, we want to expose `formNamespace` and `formPath`, not `formNamespaceAndPath`

ibacher (changes requested):

omod-2.5/pom.xml
brandones: Looks to me like it's all spaces.

brandones:
@ibacher Could you please re-review when you have the chance?

ibacher (approved):
LGTM. Thanks @IamMujuziMoses

PR #612, opened 2024-06-17
RESTWS-942 - Obs for a Concept Numeric with allowDecimal = false shou…

mseaton:
…ld not return a decimal as a value

See:  https://openmrs.atlassian.net/browse/RESTWS-942

### DIFF
--- a/omod-1.11/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_11/ObsResource1_11.java
+++ b/omod-1.11/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_11/ObsResource1_11.java
@@ -12,7 +12,11 @@ package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_11;
 import io.swagger.models.Model;
 import io.swagger.models.ModelImpl;
 import io.swagger.models.properties.StringProperty;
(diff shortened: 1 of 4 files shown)


mogoodrich (approved):
LGTM!

dkayiwa:
@mseaton could there be a reason why you did not fix this in core on the line that @chibongho pointed at?

mseaton:
> @mseaton could there be a reason why you did not just fix this in core on the line that @chibongho pointed at?

I did fix it there @dkayiwa , in a different PR associated with core that I tagged you on.  But that wouldn't fix this issue AFAIK, as the issue in core is in the getValueAsString method, whereas this REST response is the actual value itself (not necessarily as a string).

dkayiwa:
> whereas this REST response is the actual value itself (not necessarily as a string).

But the value will not have a decimal point because it was not stored with it in the first place. So `obs.getValueNumeric()` would fetch it as expected. Not so?

mseaton:
> But the value will not have a decimal point because it was not stored with it in the first place. So `obs.getValueNumeric()` would fetch it as expected. Not so?

I don't see why that would be.  It's a double in the DB.  Whether 40 or 40.0 is saved originally, the value in the db would be the same.  It's how it is read out and converted from a Double to an Integer that matters in the end, right?

dkayiwa:
Correct! 😊

PR #615, opened 2024-07-29
O3-3586 - Trying to register a patient using a duplicate identifier should block the form from submitting any data

alaboso:

This PR fixes the issue where, given a patient with a unique identifier `123456789qwerty` and we run a POST request to the `/patient/patient_uuid/identifier` endpoint, an exception is thrown `'123456789qwerty' failed to validate with reason: Identifier 123456789qwerty already in use by another patient` however, the identifier is actually persisted in the DB.

On this file, [PatientIdentifierResource1_8](https://github.com/openmrs/openmrs-module-webservices.rest/blob/master/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java) line:234, the new identifier is added to the `Set` of patient identifiers for this patient, even though it's not yet been validated. 

The next action is to call the service to save this identifier. Before this, the `RequiredDataAdvice.before` is called and through the chaining, the identifier is validated. If it fails validation, exception is thrown and response sent to the caller, BUT,  hibernate (not through the DAO) goes ahead and persists this new identifier (bypassing validation rules).

The approach is then to let the validation happen before saving this identifier. 

## Issue I worked on

https://openmrs.atlassian.net/browse/O3-3586

  No? -> execute above command

### DIFF
--- a/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java
+++ b/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java
@@ -231,10 +231,10 @@ public class PatientIdentifierResource1_8 extends DelegatingSubResource<PatientI
 		}
 		
 		if (needToAdd) {
+			service().savePatientIdentifier(delegate);
 			delegate.getPatient().addIdentifier(delegate);
 		}
-		
-		service().savePatientIdentifier(delegate);
+
 		return delegate;
 		
 	}


ojwanganto (approved):
Looks good.

dkayiwa:
@alaboso it would be great to add a test for this?

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java
```diff
@@ -231,10 +231,10 @@ public PatientIdentifier save(PatientIdentifier delegate) {
 		}
 		
 		if (needToAdd) {
+			service().savePatientIdentifier(delegate);
 			delegate.getPatient().addIdentifier(delegate);
 		}
-		
-		service().savePatientIdentifier(delegate);
```
mogoodrich, line 237: So is what happens here in the old case is the validation runs on the "savePatientIdentifier" and it is correctly flagged as invalid, but this is only *after* the identifier is attached to the patient, so it still gets saved when the patient is saved?  Bad bug and nice find!

mogoodrich (commented):
Thanks @alaboso !  See my comment, I just wanted to make sure I understand the issue properly.   I agree with @dkayiwa ... if it's relatively straightforward, would be great to add a test, both to prove that it works, and to make sure it isn't introduced in the future.

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java
alaboso: Yes.  Weird one because, the patient isn't visibly/obviously saved, except by hibernate behind the scene despite the validation exception.

alaboso:
@dkayiwa  - I have added a Test. 

- Create a unique identifier `123456789qwerty` for patient_1
- POST the same unique identifier `123456789qwerty` for patient_2
- Exception is expected AND on top of it, no new patient_identifier is created for patient_2

omod-1.9/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs1_9/PatientIdentifierController1_9Test.java
```diff
@@ -189,4 +191,37 @@ public void shouldReturnTheAuditInfoForTheFullRepresentation() throws Exception
...
+		final String OTHER_PATIENT_UUID = "5946f880-b197-400b-9caa-a3c661d23041";
+		final String REQUEST_URI = "patient/" + OTHER_PATIENT_UUID + "/identifier";
+
+		int otherPatientActiveIdentifiersSize = service.getPatientByUuid(OTHER_PATIENT_UUID).getActiveIdentifiers()
+				.size();
+		req = request(RequestMethod.POST, REQUEST_URI);
+		req.setContent(json.getBytes());
+		try {
+			handle(req);
+			fail();
```
dkayiwa, line 220: Do you mind explaining why you are using the above `fail()`?

alaboso: The idea is, If we add an identifier with a value that is already used by another patient, it should throw a validation exception. Should it not throw an exception, it means the save was a success and therefore this test should fail.

dkayiwa: Shouldn't the trowing of an exception happen automatically without having `fail()`?

alaboso: What I mean is, if `handle(req);` doesn't throw an exception, it means the `req` was successful, which it shouldn't be. 
If an exception occurs, execution won't get to `fail();` which marks this test as failed. 
The `fail();` should actually be more descriptive maybe `fail("This test fails here because the exception was not thrown yet it was expected");`

dkayiwa: Wouldn't it be clearer to just use ExpectedException

alaboso: Yes - except that in this case, After the exception is thrown, we need to check that the identifier wasn't persisted (which is the issue we were facing, that despite an exception being thrown, an identifier was actually persisted) 
@Test(expected = APIException) would pass once the exception is thrown and we wouldn't get to a point where we  test whether an identifier was persisted after the exception is thrown (like in the issue reported) or not!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientIdentifierResource1_8.java
```diff
@@ -231,10 +231,10 @@ public PatientIdentifier save(PatientIdentifier delegate) {
 		}
 		
 		if (needToAdd) {
+			service().savePatientIdentifier(delegate);
```
dkayiwa, line 234: Is there a reason why you do not start with `delegate.getPatient().addIdentifier(delegate)` before calling `service().savePatientIdentifier(delegate)`?

alaboso: Yes - This was exactly the problem.  calling `delegate.getPatient().addIdentifier(delegate)` adds this identifier to the `Set` of identifiers for this patient. At this point, this identifier has not been validated and therefore may not be valid. When `service().savePatientIdentifier(delegate)` is called, the advice triggers the validator and if the validator fails, an exception is thrown. Once this exception is thrown, hibernate still updates the `delegate.getPatient()` with the added identifier and at this point, it persists it in the db (remember it wasn't validated)

alaboso: The fix is therefore, call `service().savePatientIdentifier(delegate)` and let validation happen before OR move all this

```
if (delegate.isPreferred()) {
			for (PatientIdentifier pI : delegate.getPatient().getActiveIdentifiers()) {
				if (!pI.equals(delegate)) {
					pI.setPreferred(false);
				}
			}
		}
		
		boolean needToAdd = true;
		for (PatientIdentifier pi : delegate.getPatient().getActiveIdentifiers()) {
			if (pi.equals(delegate)) {
				needToAdd = false;
				break;
			}
		}
		
		if (needToAdd) {
			delegate.getPatient().addIdentifier(delegate);
		}
```

 logic to the openmrs-core service layer so that validation happens before these checks

ibacher (changes requested):
Because of the nature of this change we also need tests to cover:

1. Adding a new identifier works as expected
2. Modifying an existing identifier works as expected.

alaboso:
> Because of the nature of this change we also need tests to cover:
> 
> 1. Adding a new identifier works as expected
> 2. Modifying an existing identifier works as expected.

Added `shouldUpdateAnExistingPatientIdentifier` and there was an existing `shouldAddANewIdentifierToAPatient`

ibacher (approved):

gracepotma:
Asante SANA Amos for working on this!!

PR #631, opened 2024-10-14
RESTWS-958: Upgrade swagger to 2.2.23

mherman22:
## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-958

- [x] I have **added tests** to cover my changes. (If you refactored

  No? -> execute above command


omod-2.2/pom.xml
```diff
@@ -19,6 +19,28 @@
 			<groupId>${project.parent.groupId}</groupId>
 			<artifactId>${project.parent.artifactId}-omod-common</artifactId>
 			<version>${project.parent.version}</version>
+			<exclusions>
+				<exclusion>
+					<groupId>com.fasterxml.jackson.core</groupId>
```
mherman22, line 24: hacky but i needed to add these exclusions because the jackson related dependencies coming from swagger-core 2.2.23 are quite high

omod-2.3/pom.xml
```diff
@@ -19,6 +19,28 @@
             <groupId>${project.parent.groupId}</groupId>
             <artifactId>${project.parent.artifactId}-omod-common</artifactId>
             <version>${project.parent.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.core</groupId>
+                    <artifactId>jackson-annotations</artifactId>
```
mherman22, line 25: same here

omod-2.4/pom.xml
```diff
@@ -21,6 +21,28 @@
             <groupId>${project.parent.groupId}</groupId>
             <artifactId>${project.parent.artifactId}-omod-common</artifactId>
             <version>${project.parent.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.core</groupId>
```
mherman22, line 26: same here

omod-2.5/pom.xml
```diff
@@ -23,6 +23,28 @@
             <groupId>${project.parent.groupId}</groupId>
             <artifactId>${project.parent.artifactId}-omod-common</artifactId>
             <version>${project.parent.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.fasterxml.jackson.core</groupId>
```
mherman22, line 28: same here

ibacher:
@mherman22 Rather than doing the exclusions in every POM, can we add the exclusion to the `omod-common` declaration of Swagger? That should remove it from everywhere...

mherman22:
> @mherman22 Rather than doing the exclusions in every POM, can we add the exclusion to the `omod-common` declaration of Swagger? That should remove it from everywhere...

@ibacher i tried that before this approach however this bean here -> https://github.com/openmrs/openmrs-module-webservices.rest/blob/61a27fa711d211d65450b243af59a9dfff1ff4e4/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/DynamicBeanConfiguration.java#L39 causes a `org.springframework.beans.factory.BeanDefinitionStoreException` with something as 

`Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'jsonHttpMessageConverter' defined in class path resource [org/openmrs/module/webservices/rest/web/DynamicBeanConfiguration.class]: Instantiation of bean failed; nested exception is org.springframework.beans.factory.BeanDefinitionStoreException: Factory method [public org.springframework.http.converter.HttpMessageConverter org.openmrs.module.webservices.rest.web.DynamicBeanConfiguration.getMappingJacksonHttpMessageConverter() throws java.lang.Exception] threw exception; nested exception is java.lang.NoClassDefFoundError: com/fasterxml/jackson/core/JsonProcessingException`

for more content -> https://paste.ubuntu.com/p/tJ6mZxnKJJ/

ibacher:
I see what you mean... Have you tried running this version in an SDK instance?

mherman22:
> I see what you mean... Have you tried running this version in an SDK instance?

Version you mean the omod built with the changes of this PR?

If so, then no. I have an instance I setup using docker but then openmrs is failing on startup for example the queue module's 1xlegacyController class is the cause and its because i think I need to now go into each of those modules and update them to take up stuff in this PR.

However I can not do before we have this in.

Or is there a way I can actually test this on modules without having to first get this in?

ibacher:
Technically, if you run `mvn clean install` and then update the module to depend on the SNAPSHOT version, you can test this all locally.

ibacher:
`install` adds it to your local Maven repo copy and, at least until the next PR gets merged, that should be newer than the SNAPSHOT we've published (and you can always re-do the install if that happens), so Maven will pick-up your local copy as the "latest" SNAPSHOT.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerConstants.java
mherman22, line 1: pulled this out of RestConstants because for soe reason, initialization of RestConstants was failing.

omod-common/src/main/java/org/openmrs/module/webservices/docs/swagger/SwaggerSpecificationCreator.java
```diff
@@ -140,19 +156,20 @@ private void BuildJSON() {
...
-		} else {
-			openAPI = new OpenAPI();
-			BuildJSON();
+			return cachedJson;
 		}
-		return createJSON();
+
+		openAPI = new OpenAPI();
+		BuildJSON();
+		cachedJson = createJSON();  // Cache the JSON string
```
mherman22, line 171: The Spec was being reset when accessing cached data, causing
empty paths in the API documentation on subsequent requests. This occurred
because `initOpenAPI()` was creating a new empty Paths object while handling
cached responses.

mherman22:
Screenshot 2024-10-29 at 14-49-24 OpenMRS
these are the current visuals. I am lookng into the failing test

omod-common/pom.xml
```diff
@@ -31,9 +31,9 @@
 		</dependency>
 
 		<dependency>
-			<groupId>io.swagger</groupId>
+			<groupId>io.swagger.core.v3</groupId>
```
ibacher, line 34: Quick thing: since the versions are already defined in the root POM (in dependency management), they don't need to be specified here. That way we only need to update versions in a single place (dependency management doesn't make something a dependency, but it's configuration applies to a dependency with the same coordinates.

For the same reasons, we shouldn't need to replicate the exclusions here.

mherman22: fixed

dkayiwa:
@mherman22 as we discussed on this week's platform call, did you get some time to check if swagger has an option that saves us from the burden of having to manually maintain resource documentation?

mherman22:
> @mherman22 as we discussed on this week's platform call, did you get some time to check if swagger has an option that saves us from the burden of having to manually maintain resource documentation?

@dkayiwa I am still looking into that. But so far what i have landed on from the various docs that i have read indicate OpenAPI 3.0 itself may not provide a mechanism to automatically generate schema definitions from code. But i am looking into [Springdoc](https://springdoc.org/) trying to see if it works for us but it looks to be more of a springboot project.

Others that i have looked at are poorly maintained and it's been long (4 years) since the last commit.

I will take any suggestions before i read all books on the internet 😃

mherman22:
add [springfox](https://www.baeldung.com/swagger-2-documentation-for-spring-rest-api) to that list.

dkayiwa:
@mherman22 do you happen to know why our swagger docs no longer load as they used to?
In the past, we at least had something like this: https://o2.openmrs.org/openmrs/module/webservices/rest/apiDocs.htm

But now, not any more as you can see here: https://dev3.openmrs.org/openmrs/module/webservices/rest/apiDocs.htm

mherman22:
> @mherman22 do you happen to know why our swagger docs no longer load as they used to? In the past, we at least had something like this: https://o2.openmrs.org/openmrs/module/webservices/rest/apiDocs.htm
> 
> But now, not any more as you can see here: https://dev3.openmrs.org/openmrs/module/webservices/rest/apiDocs.htm

@dkayiwa on dev3, it looks like only a bit of documentation from modules is available, and none from the parent rest-module. Let me reproduce that first locally. Usually [these tests](https://github.com/openmrs/openmrs-module-webservices.rest/blob/e81036bf82e2023454a004d10500147fad071b5c/omod-2.0/src/test/java/org/openmrs/module/webservices/rest/doc/SwaggerSpecificationCreatorTest.java#L279-L303) should fail

mherman22:
@dkayiwa i was not able to reproduce this locally but i did experience it sometime back while working on one of these pull requests.

Restarting the server should work. The spec generation is very unstable.

dkayiwa:
Yes restarting made it work again. Very unstable!

vickabire:
> Yes restarting made it work again. Very unstable!

Not sure whether to 😄 or 😢

PR #642, opened 2025-03-14
RESTWS-967: Return Obs of type "Date" in "yyyy-MM-dd" format

mogoodrich:
## Issue I worked on

[see https://issues.openmrs.org/browse/RESTWS-](https://openmrs.atlassian.net/browse/RESTWS-967)

  No? -> execute above command

### DIFF
--- a/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
+++ b/omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
@@ -12,6 +12,7 @@ package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -71,7 +72,11 @@ import io.swagger.models.properties.StringProperty;
  */
 @Resource(name = RestConstants.VERSION_1 + "/obs", order = 2, supportedClass = Obs.class, supportedOpenmrsVersions = { "1.8.*" })
 public class ObsResource1_8 extends DataDelegatingCrudResource<Obs> implements Uploadable {
-	
+
+	public static final String DATETIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+
+	public static final String DATE_FORMAT = "yyyy-MM-dd";
+
 	/**
 	 * @see org.openmrs.module.webservices.rest.web.resource.impl.BaseDelegatingResource#delete(java.lang.Object,
 	 *      java.lang.String, org.openmrs.module.webservices.rest.web.RequestContext)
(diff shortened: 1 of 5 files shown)


omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
```diff
@@ -296,7 +301,10 @@ public Object getValue(Obs obs) throws ConversionException {
 			return null;
 		
 		if (obs.getValueDatetime() != null) {
```
mogoodrich, line 303: So this changes this getValue method to return a String instead of Date for obs of time "date" and "datetime". Not sure how correct this is, as the getValue method return signature is "Object", and it generally returns Objects (if you take a look at the full code), relying on downstream converters to convert to a string, *but* the method signature does say its "for converting a value to a string", and, at the end of the day, that is what we want a REST resource handlers to do, right (any REST response is just a JSON collection of primitives)?

If we wanted to keep this at the object level, the correct thing to do IMO would be to change the below to return a LocalDate if type "isDate" and java "Date" if otherwise, and change the ConversionUtil to properly parse a LocalDate to "yyyy-MM-dd" (which is what the local date toString method returns anyway).  I started doing this but it was a rabbit hole because REST Web Services still supports back to OpenMRS Core 1.11 (and therefore builds on Java 7) and there would be a lot of cleanup to do to fix it... though this is something we are definitely going to need to do in the not-so-distance future anyway; if we wanted we could argue that we should wait to make these changes until we do that, and I could ticket that.

Another option here would be to defer to the Core Obs getValueAsString(Locale) method and then fix that in Core... this is what ObsTree uses and it *does* do the correct thing and returns "yyyy-MM-dd" for Obs of datatype=Date, but returns a ISO string without a timezone for DateTime  (which is the original problem we were facing that lead to this ticket).  Although this would be nice to do, I think with slow release cycle of Core this might not be practical.  We'd also need think how the Locale component of "getValueAsString(Locale)" would/does play into this.  (Maybe not a big deal, but I haven't given it any thought).

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
```diff
@@ -296,7 +301,10 @@ public Object getValue(Obs obs) throws ConversionException {
...
-			return ConversionUtil.convert(obs.getValueDatetime(), Date.class);
```
mogoodrich, line 299: FWIW I think this line wasn't doing anything, basically saying "convert an Object of type Date to an object of type Date".

mogoodrich (commented):
Still a draft, but @ibacher  @dkayiwa  @mseaton  I wanted to out this out there for your review before going farther.  See my long-winded comment on the PR.  :)

Before marking as ready for merging I'd at least want to also add tests (scarily but not surprisingly, this change as-is doesn't break any tests).  I'd also update ObsTree to use the same logic, and would want to do some manual "round trip" testing of submitting an Obs of type Date and getting it back.

Anyway, thanks!

omod-1.8/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
dkayiwa: I do not see any reason why we cannot change the required version from OpenMRS Core 1.11 to something like 2.x

mogoodrich: Agree 100% about updating REST Web Services to some 2.x version sooner rather than later, but with all the the legacy 1.x submodules, etc, this was going to be nontrivial so I didn't want to go down to that rabbit hole.

I went ahead and ticketed that here: https://openmrs.atlassian.net/browse/RESTWS-968

ibacher (approved):

PR #655, opened 2025-05-07
RESTWS-979 - Custom representations are not parsed correctly

mseaton:
see https://issues.openmrs.org/browse/RESTWS-979

Note, in addition to the unit test included here which demonstrates the issue, I have verified that this also fixes the test that I was attempting to write in EMR API that first caused me to investigate this issue.

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
@@ -547,7 +547,7 @@ public class ConversionUtil {
 		int startIndex = 0;
 		List<String> properties = new ArrayList<String>();
 		int nestingLevel = 0;
-		for (int i=0; i<def.length(); i++) {
+		for (int i=0; i < def.length(); i++) {
 			char c = def.charAt(i);
 			if (c == '(') {
 				nestingLevel++;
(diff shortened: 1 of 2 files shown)


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/ConversionUtil.java
```diff
@@ -539,56 +543,46 @@ public static DelegatingResourceDescription getCustomRepresentationDescription(C
 		String def = representation.getRepresentation();
 		def = def.startsWith("(") ? def.substring(1) : def;
 		def = def.endsWith(")") ? def.substring(0, def.length() - 1) : def;
```
mseaton, line 545: I don't know where this original code came from - I am shown on the blame in github, but I have no recollection writing this and it doesn't look like my style, so I probably moved it from somewhere and that history is lost.

In any event, this seems to work in some cases (obviously it must), but does not work for certain combinations of nested properties.  As can be seen below, the original parsing was pretty brittle, splitting on commas even though commas can be used to separate fields both at the top level and within nested objects, and the code is not resilient to all cases.

ibacher (approved):
LGTM

dkayiwa (approved):

dkayiwa:
Very interesting that for all these years we have never go a problem with this. 😊

mogoodrich (approved):

PR #672, opened 2025-09-04
RESTWS-990: Allow audit data to be creatable

mogoodrich:
## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-990

  No? -> execute above command


omod-2.4/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientResource1_8.java
```diff
@@ -195,6 +195,18 @@ public DelegatingResourceDescription getCreatableProperties() {
...
+			description.addProperty("creator");
+			description.addProperty("dateCreated");
+			description.addProperty("changedBy");
+			description.addProperty("dateChanged");
+			description.addProperty("personDateCreated");
+			description.addProperty("personCreator");
+			description.addProperty("personDateChanged");
+			description.addProperty("personChangedBy");
+		}
+
```
rkorytkowski: Anything stopping you to apply it generally to all resources in DelegatingCrudResource and DelegatingSubResource for objects implementing Auditable/Retireable/Voidable?

ibacher: Good idea.

@mogoodrich What if we approached this from a slightly different angle and instead of modifying _every_ resource we just patched in something [here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/e8838aae8cd701aa789b2eca8d6531e963bd8eef/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java#L103) and [here](https://github.com/openmrs/openmrs-module-webservices.rest/blob/e8838aae8cd701aa789b2eca8d6531e963bd8eef/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java#L137) based on whether the class extended `Auditable` (we actually do something similar for Voidable / Retireable, but only on updates; see the second link)? In both those cases, we actually have an instance of the object created by the `DelegatingCrudResource` so type erasure wouldn't be an issue.

That would leave the only "weird" case as the `Patient` itself, because we have the non-standard `personDateChanged` fields, which we'd presumably want to expose as well?

I guess the issue is that these wouldn't show up in the declared full representation (which maybe doesn't matter) or in the Model (so they wouldn't show up in the Swagger docs), but maybe those are separately patchable?

mogoodrich: @rkorytkowski  @ibacher thats certainly easier on my end, I just didn't know if there's be appetite to do that because it's kind of hidden....and somehow I missed the precedent where we added voided and retired in the getUpdatable method.

I can add the person case directly to Patient. I will move forward with that, thanks!

omod-2.4/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/PatientResource1_8.java
```diff
@@ -195,6 +195,30 @@ public DelegatingResourceDescription getCreatableProperties() {
 		DelegatingResourceDescription description = new DelegatingResourceDescription();
 		description.addRequiredProperty("person");
 		description.addRequiredProperty("identifiers");
+
+		// we apply these generically to resources in the DelegationCrudResource before calling setConvertedProperties,
+		// but the PatientResource calls setConvertedProperties directly in getPatient method below, so we need to add this here
```
mogoodrich: fyi @rkorytkowski @ibacher as to why we need to apply all the audit properties here on Patient

omod-2.4/src/test/java/org/openmrs/module/webservices/rest/web/DelegatingCrudResourceTest.java
```diff
@@ -172,5 +175,74 @@ public void convert_shouldConvertASimpleObjectThatIncludesAUuid() {
 		Location location = resource.convert(so);
 		Assert.assertEquals(uuid, location.getUuid());
 	}
-	
+
```
mogoodrich: FWIW, I was forgetting if there was a straightforward way to test with a user that *does not* have the set audit data privilege.

PR #723, opened 2026-03-22
RESTWS-971: Fix property double-set in convertMap causing PatientProgram state transition error

Binayak490-cyber:
## Problem

When a `PatientProgram` is submitted as a nested property inside a wrapper resource, `ConversionUtil.convertMap()` sets the `states` property **twice** on the same object. The second call triggers OpenMRS's state transition guard — *"You can't change from state X to state X"* — causing the API to reject a perfectly valid request with a misleading error. No data gets saved.

## Root Cause

`convertMap()` has two loops that both call `setProperty()` with no deduplication between them. A fix was previously attempted in RESTWS-901 (adding `alreadySetProperties`) but was reverted in RESTWS-933 because coded patient attributes broke.

The reason they broke: `PersonAttributeResource1_8` declared `value` **before** `attributeType` in `DefaultRepresentation`. So Loop 1 set `value` with a null `attributeType`, and Loop 2 had to re-set it correctly after `attributeType` was set. The `alreadySetProperties` fix prevented Loop 2 from re-setting `value`, breaking coded attribute conversion.

## Fix

* **`PersonAttributeResource1_8`** — Move `attributeType` before `value` in `DefaultRepresentation` so Loop 1 processes them in the correct order
* **`ConversionUtil.convertMap()`** — Restore `alreadySetProperties` deduplication so Loop 2 skips properties already handled by Loop 1

With both changes together:

* `PersonAttribute` coded values still convert correctly (attributeType is set before value in Loop 1)
* `PatientProgram.states` is set exactly once (Loop 2 skips it)

## Tests

* `convertMap_shouldCallSetPropertyOnlyOnceForPropertiesInDefaultRepresentation` — proves `setProperty` is called exactly once for properties in `DefaultRepresentation`
* `create_shouldCorrectlyConvertCodedAttributeValueWhenCreatedViaConvertMap` — proves coded patient attributes still convert correctly, directly covering the RESTWS-933 revert scenario

## Results

All existing tests pass with zero failures.


Binayak490-cyber:
@ibacher, Closing this for a new PR - #724 which contains only the RESTWS-971 fix on top of the latest upstream master.

PR #726, opened 2026-03-30
feat(RESTWS-1131): add PropertyDecorator support for extending REST resources

vivektarun:
## Summary

Adds `PropertyDecorator` support so that modules can add extra properties to existing REST resources without subclassing them.

Currently, if a module wants to expose an additional property on an existing resource (e.g. adding `fulfillerEncounter` to the Order resource), it must subclass the existing resource class. This is fragile — the subclass breaks whenever a newer version of the resource is released.

This PR introduces a `PropertyDecorator` marker interface that allows any Spring component to register extra properties for a domain object without touching the existing resource class.

## Changes Made

- Added `PropertyDecorator.java` — a marker interface that   modules implement to register as a property decorator
- Updated `BaseDelegatingConverter.getProperty()` — after  checking the resource class for `@PropertyGetter` now also checks registered `PropertyDecorator` beans via Spring context
- Added `ReflectionUtil.findPropertyGetterMethodOnObject()` —   a helper method to find `@PropertyGetter` on any plain object, not just `DelegatingPropertyAccessor` instances

## How to Use
```java
@Component
@Handler(supports = Order.class)
public class MyOrderDecorator implements PropertyDecorator {

    @PropertyGetter("fulfillerEncounter")
    public Encounter getFulfillerEncounter(Order order) {
        return ...;
    }
}
```

## Screenshots

N/A — backend only change

## Related Issue

Fixes https://openmrs.atlassian.net/browse/RESTWS-1021

## Other

The build failure in `TaskServiceWrapper.java` is a pre-existing issue unrelated to this PR. Confirmed by reproducing the same error on the unmodified master branch before applying any changes.

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingConverter.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingConverter.java
@@ -28,6 +28,9 @@ import org.openmrs.module.webservices.rest.web.representation.CustomRepresentati
 import org.openmrs.module.webservices.rest.web.representation.Representation;
 import org.openmrs.module.webservices.rest.web.resource.api.Converter;
 import org.openmrs.module.webservices.rest.web.response.ConversionException;
+import org.openmrs.annotation.Handler;
+import org.openmrs.api.context.Context;
+import org.openmrs.module.webservices.rest.web.resource.impl.PropertyDecorator;
 
 /**
  * A base implementation of a converter that can transform something that is _not_ a full resource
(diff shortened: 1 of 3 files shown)


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingConverter.java
```diff
@@ -74,21 +77,65 @@ public SimpleObject asRepresentation(T delegate, Representation rep) throws Conv
 
 		return convertDelegateToRepresentation(delegate, description);
 	}
-	
+
+	// AFTER
```
Sachindu-Nethmin:
```suggestion
```

Sachindu-Nethmin: Could we consider removing the // AFTER comment? It appears to be a
   merge/rebase artifact and doesn't add clarity. This would keep the code
   cleaner for future readers.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/PropertyDecorator.java
```diff
@@ -0,0 +1,14 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest.web.resource.impl;
+
```
Sachindu-Nethmin, line 11: Should we add javadoc to explain the interface's purpose? It would
  help future developers understand that this is a marker interface and
  that implementations need the @Handler annotation. This seems like it
  would be valuable documentation.
```suggestion

/**
   * Marker interface for property decorators that provide additional
   * property getters for domain objects. Decorators must be annotated
   * with @Handler to specify the domain classes they support.
   */
```

PR #729, opened 2026-04-13
Separate OWASP dependency check into its own job

ibacher:
## Summary
- Separate OWASP dependency check into a standalone workflow job
- OWASP check only runs on push and workflow_dispatch (not on PRs)
- Build job no longer runs the OWASP check inline or needs the NVD_API_KEY secret

## Test plan
- [ ] Verify build job still passes
- [ ] Verify OWASP dependency check runs on push to the default branch
- [ ] Verify OWASP dependency check does not run on pull requests


ibacher:
Failures here are due to changes to the scheduler that are not implemented in this module yet.

PR #731, opened 2026-04-15
RESTWS-1027: Fix compilation error in TaskServiceWrapper

chibongho:

This PR fixes compilation error related to the SchedulerService changes in `openmrs-core` [(PR)](https://github.com/openmrs/openmrs-core/pull/5893). (The master branch of RESTWS is the only branch affected by this)

I have not tested this. `mvn clean install -DskipTests` passes. If I run the tests, several tests fail, but are unrelated to this change.  

## Issue I worked on

https://openmrs.atlassian.net/browse/RESTWS-1027

  No? -> execute above command

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/helper/TaskServiceWrapper.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/helper/TaskServiceWrapper.java
@@ -11,13 +11,10 @@
 package org.openmrs.module.webservices.helper;
 
 import org.openmrs.scheduler.TaskDefinition;
-import org.openmrs.scheduler.TaskFactory;
 import org.openmrs.scheduler.SchedulerException;
-import org.openmrs.scheduler.Task;
 import org.openmrs.api.context.Context;
 
 import java.util.Collection;
-import java.util.concurrent.ExecutionException;
 
 /**
  * Wrapping the required task information between related resource and SchedulerService
@@ -106,11 +103,6 @@ public class TaskServiceWrapper {
 	 * @throws SchedulerException - It will throw in case of any SchedulerService exceptions
 	 */
 	public void runTask(TaskDefinition taskDefinition) throws SchedulerException {
-		Task task = TaskFactory.getInstance().createInstance(taskDefinition);
-        try {
-            task.execute();
-        } catch (InterruptedException | ExecutionException e) {
-            throw new RuntimeException(e);
-        }
-    }
+		Context.getSchedulerService().schedule(taskDefinition);
+	}
 }


mseaton (commented):
Is there a ticket for this?  What exactly is this used for?  Why do we have tasks in the webservices module?

ibacher:
> Why do we have tasks in the webservices module?

Chi Bong is just trying to fix the build failures. The functionality to manage tasks via REST was added in 2017 via RESTWS-673, but platform 3 and 2.9 have the new scheduling system which breaks API compatibility, hence build failures.

mogoodrich:
If we are just trying to fix the build failures, I agree with @ibacher that we don't need to reassess (as part of this fix) whether we should really have scheduled tasks as part of the REST WS module... *but* I would agree with @mseaton  that we should at least have a quick ticket around this with a few details so we can track/remember why we made this change in the future. @chibongho

chibongho:
Added tracking ticket: https://openmrs.atlassian.net/browse/RESTWS-1027

omod-common/src/main/java/org/openmrs/module/webservices/helper/TaskServiceWrapper.java
```diff
@@ -106,11 +103,6 @@ public void reScheduleAllTasks() throws SchedulerException {
...
 	public void runTask(TaskDefinition taskDefinition) throws SchedulerException {
-		Task task = TaskFactory.getInstance().createInstance(taskDefinition);
-        try {
-            task.execute();
-        } catch (InterruptedException | ExecutionException e) {
-            throw new RuntimeException(e);
-        }
-    }
+		Context.getSchedulerService().schedule(taskDefinition);
+	}
```
mogoodrich, line 107: @rkorytkowski just wanted to confirm, should this be backwards compatible with versions of Core prior to this change?  fyi @ibacher 

@chibongho  are there any tests around this or can we dig in to see how this is actually used (so we can test IRL if things work with 2.8.x).  

And, also, if rework does turn out to be complicated can we consider per @mseaton  if this functionality is actually needed--I haven't done any research about how this is actually used.

mseaton: Thanks @mogoodrich , these were the points/questions I was trying to raise

rkorytkowski: `Context.getSchedulerService().schedule(taskDefinition);` should work in all versions of core. It's a change of behaviour for runTask. It's only scheduled and not executed right away in the same thread as with task.execute() so we do need a jira issue to address this properly if really needed... I'd be surprised if the old implementation was intentional.

mogoodrich: Thanks @rkorytkowski  for this insight, I dug into this a bit more and this change is indeed incorrect... the method is to support triggering a Task to run one-time RESTfully, see https://openmrs.atlassian.net/browse/RESTWS-761 and the linked Talk post.  (You can see there is already an existing way to actually schedule a task).

@rkorytkowski  it sounds like there is no longer a way to run a task one-time?

mogoodrich (changes requested):
This needs rework, see my comments...

omod-common/src/main/java/org/openmrs/module/webservices/helper/TaskServiceWrapper.java
rkorytkowski: @mogoodrich Yes, there's. In the new API you can call schedule(...) to have one-off task scheduled or scheduleRecurrently(...) to have a recurring task. If you want to use the legacy TaskDefinition then tasks with repeatInterval set to null are one-off tasks. You may set startTime to null as well to have tasks run as soon as possible. You need to remember that changes to TaskDefinition will be persisted when calling schedule(TaskDefintion) so if you want the task definition to continue to be recurring after a one-off run, then you need to reschedule the task after the one-off run with repeatInterval set.

mogoodrich:
It seems like this not the correct fix, but we can't leave a much-used module like this with compile errors for this long.  I'm going to close this PR, apply a fix just to disable this feature so compiling will work, and have created a new ticket for fixing this, see: https://openmrs.atlassian.net/browse/RESTWS-761?search_id=5944be95-cd94-4471-936b-6755ead9b985

fyi @rkorytkowski @chibongho @mseaton  @rkorytkowski

chibongho:
I dug into this more. I actually think this is the correct fix. `Context.getSchedulerService().schedule(taskDefinition)` calls JobRunr's `jobRequestScheduler.enqueue(job)`, which schedule the task for a "fire-and-forget" job (See [javadocs](https://repo.jobrunr.io/javadoc/releases/org/jobrunr/jobrunr-pro/latest/.cache/unpack/org/jobrunr/scheduling/JobRequestScheduler.html#enqueue(org.jobrunr.jobs.lambdas.JobRequest)). That said, it's not an immediate run of the job. By default, JobRunr has a polling interval of 15 seconds, See [here](https://javadoc.io/static/org.jobrunr/jobrunr/5.3.2/org/jobrunr/server/BackgroundJobServerConfiguration.html#usingStandardBackgroundJobServerConfiguration()). It doesn't quite match the old behavior of firing right away, but given that it's a scheduled job, it's probably acceptable?

I also found existing tests that were testing this TaskServiceWrapper, and were failing before. I have updated the PR to fix them. 

2 more tests are still failing, but un-related to this:
```
[ERROR] Errors: 
[ERROR]   CreatePatientIdentifierResource1_9Test.shouldCreatePatientIdentifier_WhenTypeIsSpecifiedByName:62 » Validation '101-6' failed to validate with reason: PatientIdentifier.error.notUniqueSamePatientWithParameter
[ERROR]   CreatePatientIdentifierResource1_9Test.shouldCreatePatientIdentifier_WhenTypeIsSpecifiedByUuid:47 » Validation '101-6' failed to validate with reason: PatientIdentifier.error.notUniqueSamePatientWithParameter

``` 

Re-opened PR here: https://github.com/openmrs/openmrs-module-webservices.rest/pull/736.

rkorytkowski:
@chibongho Context.getSchedulerService().schedule(taskDefinition) enqueues or schedules recurrently depending on what's defined in the taskDefinition. My understanding is that the point of this REST call is to run the task right away regardless if it is configured to run right away or recurrently. I think it's acceptable to run it "almost" right away that is using enqueue with the scheduler service instead of in the same thread as it was implemented originally. In order to achieve that you need to modify taskDefinition as described above.

chibongho:
@rkorytkowski `Context.getSchedulerService().schedule(taskDefinition)` actually enqueues the task to run almost immediately and just once, ignoring the `taskDefinition`'s parameters like `startTime` or ``repeatInterval`. I updated the test in https://github.com/openmrs/openmrs-module-webservices.rest/pull/736 to verify that.

rkorytkowski:
Right! I thought you were calling the deprecated scheduler method, which is called scheduleTask(TaskDefinition) and not schedule(TaskDetails).

PR #740, opened 2026-05-07
(fix) RESTWS-1035: Deduplicate concept results in memberOf and fuzzy search paths

sohamdhande:

The `doSearch()` method in `ConceptResource1_8.java` and the fuzzy search path in `ConceptSearchHandler1_8.java` never deduplicated their results lists. Both `getConceptsByConceptSet()` and `getConcepts()` can return duplicate `Concept` objects due to Hibernate JOIN FETCH on multi-valued associations (`concept_name`, `concept_map`), causing the REST response to contain the same concept multiple times.

### Root Cause
- `getConceptsByConceptSet()` uses an HQL query that joins on the `concept_set` table and can produce duplicate rows when a concept has multiple names or mappings
- `getConcepts()` similarly produces duplicates from joins on the `concept_name` table
- Neither the `memberOf`, `answerTo`, nor `q` search paths ever deduplicated the final results

### Changes Made

| File | Change |
|------|--------|
| `ConceptResource1_8.java` | Added `LinkedHashSet` deduplication after results list is built, covering `memberOf`, `answerTo`, and `q` search paths |
| `ConceptSearchHandler1_8.java` | Added the same `LinkedHashSet` dedup in the `fuzzy` searchType path |
| `ConceptController1_8Test.java` | Added `doSearch_shouldNotReturnDuplicateMembersOfConcept()` to assert no duplicate UUIDs are returned when querying `memberOf` |

### Why LinkedHashSet?
`LinkedHashSet` eliminates duplicates while preserving insertion order, making it the safest and most localized fix no upstream query behavior is changed.

## Issue I worked on
see https://issues.openmrs.org/browse/RESTWS-1035

- [x] I have **added tests** to cover my changes.

### DIFF
--- a/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ConceptResource1_8.java
+++ b/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ConceptResource1_8.java
@@ -61,6 +61,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Locale;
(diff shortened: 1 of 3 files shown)


sohamdhande:
cc: @jayasanka-sack @ibacher

ibacher:
This feels like a fix that belongs in the core API and not in the REST layer to me. Generally speaking the REST API should be a REST representation of the core API and should not implement it's own functionality. There are a handful of small exceptions where we need some functionality exposed via REST.

PR #751, opened 2026-06-13
RESTWS-1038: Include obs uuid in the obstree response

solomonfortune:

The `ObsTreeResource1_9` (used by the `/ws/rest/v1/obstree` endpoint) builds a map for each observation in the tree containing `value`, `obsDatetime`, `interpretation`, and reference range fields, but did not include the observation's own `uuid`.

Without the `uuid`, clients (specifically the patient chart's test results viewer) cannot reliably distinguish:
- The same observation appearing in multiple branches of the tree (e.g. when configured root concept sets overlap, such as "Bloodwork" containing "Hematology"), from
- Two genuinely distinct observations that happen to share the same value and `obsDatetime` (e.g. one result per order, entered against the same encounter).

The frontend currently de-duplicates using an `obsDatetime` + `value` heuristic, which cannot tell these two cases apart.

This PR adds `obs.getUuid()` to the observation map in `ObsTreeResource1_9#getObsMap`, alongside the existing `value` and `obsDatetime` fields.

This change is additive and backwards compatible: existing consumers can simply ignore the new `uuid` field, while the frontend can switch its de-duplication identity to the obs `uuid` when present, falling back to its existing heuristic for older servers.

Related frontend PR: https://github.com/openmrs/openmrs-esm-patient-chart/pull/3374

## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-1038

- [x] I have **added tests** to cover my changes. (Updated the expected fixture `obsTreeDataset.json` used by `ObsTreeController1_9Test#shouldGetObsTree` to assert the `uuid` field is present and correct for each observation.)
- [x] All new and existing **tests passed**. (`ObsTreeController1_9Test` passes; remaining unrelated failures on the full suite — `TaskActionController*`, `VisitConfigurationController2_0Test`, `CreatePatientIdentifierResource1_9Test` — are pre-existing and unaffected by this change.)

### DIFF
--- a/omod/src/test/resources/obsTreeDataset.json
+++ b/omod/src/test/resources/obsTreeDataset.json
@@ -10,7 +10,8 @@
           "value" : "7.0",
           "interpretation" : "NORMAL",
           "hiAbsolute" : 60.0,
-          "hiCritical" : 68.0
+          "hiCritical" : 68.0,
+          "uuid" : "a1521c32-47b6-47da-9c6f-3673ddfb74f1"
         } ],
         "datatype" : "Numeric",
         "display" : "Hematocrit",
@@ -21,7 +22,8 @@
       }, {
         "obs" : [ {
           "obsDatetime" : "2008-08-15T00:00:00.000+0200",
-          "value" : "8.0"
+          "value" : "8.0",
+          "uuid" : "b2521c32-47b6-47da-9c6f-3673ddfb74f1"
         } ],
         "datatype" : "Numeric",
         "display" : "Platelets",
@@ -32,7 +34,8 @@
       }, {
         "obs" : [ {
           "obsDatetime" : "2008-08-15T00:00:00.000+0200",
-          "value" : "9.0"
+          "value" : "9.0",
+          "uuid" : "c3521c32-47b6-47da-9c6f-3673ddfb74f1"
         } ],
         "datatype" : "Numeric",
         "display" : "Haemoglobin",
@@ -43,7 +46,8 @@
       }, {
         "obs" : [ {
           "obsDatetime" : "2008-08-15T00:00:00.000+0200",
-          "value" : "10.0"
+          "value" : "10.0",
+          "uuid" : "d4521c32-47b6-47da-9c6f-3673ddfb74f1"
         } ],
         "datatype" : "Numeric",
         "display" : "White blood cells",
@@ -60,7 +64,8 @@
       "subSets" : [ {
         "obs" : [ {
           "obsDatetime" : "2008-08-15T00:00:00.000+0200",
-          "value" : "62.0"
+          "value" : "62.0",
+          "uuid" : "aa521c32-47b6-47da-9c6f-3673ddfb74f1"
         } ],
         "datatype" : "Numeric",
         "display" : "Serum Chloride",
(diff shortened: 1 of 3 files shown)


omod/src/test/resources/obsTreeDataset.json
```diff
@@ -141,7 +153,8 @@
             "subSets" : [ {
               "obs" : [ {
                 "obsDatetime" : "2008-08-15T00:00:00.000+0200",
-                "value" : ""
+                "value" : "",
+                "uuid" : "06521c32-47b6-47da-9c6f-3673ddfb74f1"
```
IamMujuziMoses, line 157: Have you tried running this in postman, what response do you get.

omod/src/test/resources/obsTreeDataset.json
```diff
@@ -10,7 +10,8 @@
           "value" : "7.0",
           "interpretation" : "NORMAL",
           "hiAbsolute" : 60.0,
-          "hiCritical" : 68.0
+          "hiCritical" : 68.0,
+          "uuid" : "a1521c32-47b6-47da-9c6f-3673ddfb74f1"
```
IamMujuziMoses, line 14: Hard coded uuids are not optimal

IamMujuziMoses (commented):
Thank you for this contribution @solomonfortune I have left some comments below.

omod/src/test/resources/obsTreeDataset.json
solomonfortune: I attempted to test this via a local OpenMRS SDK server but ran into environment issues during server setup (SPA build failure unrelated to this change). The change itself is a single additive line i.e.valueMap.put("uuid", obs.getUuid()) in ObsTreeResource1_9#getObsMap. The uuid field is sourced directly from obs.getUuid() which is inherited from BaseOpenmrsObject and is always populated for persisted obs.

omod/src/test/resources/obsTreeDataset.json
solomonfortune: @IamMujuziMoses I think these UUIDs are not arbitrary hardcoded values because they are taken directly from the uuid column of the obs rows defined in obsTreeDataset.xml

IamMujuziMoses: > they are taken directly from the uuid column of the obs rows defined in obsTreeDataset.xml

`obsTreeDataset.xml` is used in tests, they are not real uuid values, real uuid values are retrieved from the database.

solomonfortune: Thank you for the clarification. I kind of understand the concern asserting on specific UUID string literals in the fixture is brittle. I stand to be corrected. Would you prefer me to update the test to dynamically assert that a uuid key is present and non-null in each obs entry, rather than comparing against hardcoded values in the JSON fixture?
Or would you prefer a different approach entirely? Happy to follow whatever pattern the codebase prefers.

IamMujuziMoses: > Would you prefer me to update the test to dynamically assert that a uuid key is present and non-null in each obs entry, rather than comparing against hardcoded values in the JSON fixture?

Yes, this would be the right approach

solomonfortune: Updated as suggested. The test now asserts that each obs entry has a non-null, non-empty uuid dynamically rather than comparing against hardcoded uuid values in the JSON fixture. By the way, the fixture itself no longer contains uuid values.

PR #753, opened 2026-07-19
RESTWS-1039: Support viewing archived observations introduced in TRUNK-6649

shivvani-r:
This PR resolves an issue following the introduction of observation archiving in TRUNK-6649, where archived observations were missing from the REST API payload when querying an encounter with `includeAll=true` (e.g., when a user clicks "Toggle Deleted" in the UI).

## Specific changes include

- `ObsResource1_8.java`: Updated the resource logic to explicitly call the newly introduced `Encounter.getAllObsIncludingArchived()` method when includeAll is requested, instead of `Encounter.getAllObs()`. This ensures archived observations are successfully returned to the client.

## Issue I worked on
see https://issues.openmrs.org/browse/RESTWS-1039

  No? -> execute above command

### DIFF
--- a/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
+++ b/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
@@ -574,7 +574,12 @@ public class ObsResource1_8 extends DataDelegatingCrudResource<Obs> implements U
 			if (enc == null)
 				return new EmptySearchResult();
 			
-			List<Obs> obs = new ArrayList<Obs>(enc.getAllObs(context.getIncludeAll()));
+			List<Obs> obs;
+			if (context.getIncludeAll()) {
+				obs = new ArrayList<Obs>(enc.getAllObsIncludingArchived());
+			} else {
+				obs = new ArrayList<Obs>(enc.getAllObs(false));
+			}
 			return new NeedsPaging<Obs>(obs, context);
 		}
 		
--- a/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
+++ b/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
@@ -11,10 +11,14 @@ package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import org.openmrs.api.APIException;
+import org.openmrs.api.impl.ObsArchiveHelper;
+import org.openmrs.util.OpenmrsConstants;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -378,10 +382,23 @@ public class ObsResource1_8 extends DataDelegatingCrudResource<Obs> implements U
 	 */
 	@PropertyGetter("groupMembers")
 	public static Object getGroupMembers(Obs obs) throws ConversionException {
-		if (obs.getGroupMembers() != null && !obs.getGroupMembers().isEmpty()) {
-			return obs.getGroupMembers();
+		Set<Obs> members = obs.getGroupMembers();
+		if (members == null) {
+			members = new LinkedHashSet<Obs>();
+		} else {
+			members = new LinkedHashSet<Obs>(members);
 		}
-		return null;
+
+		// Also include any archived children from obs_archive
+		ObsArchiveHelper archiveHelper = getObsArchiveHelperSafely();
+		if (archiveHelper != null && obs.getObsId() != null) {
+			List<Obs> archivedChildren = archiveHelper.getArchivedChildObs(obs.getObsId());
+			if (archivedChildren != null) {
+				members.addAll(archivedChildren);
+			}
+		}
+
+		return members.isEmpty() ? null : members;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


shivvani-r:
cc: @dkayiwa @Muta-Jonathan

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
```diff
@@ -574,7 +591,12 @@ protected PageableResult doSearch(RequestContext context) {
 			if (enc == null)
 				return new EmptySearchResult();
 			
-			List<Obs> obs = new ArrayList<Obs>(enc.getAllObs(context.getIncludeAll()));
+			List<Obs> obs;
+			if (context.getIncludeAll()) {
+				obs = new ArrayList<Obs>(enc.getAllObsIncludingArchived());
```
dkayiwa, line 596: `Encounter.getAllObsIncludingArchived()`, `org.openmrs.api.impl.ObsArchiveHelper` and `OpenmrsConstants.GP_OBS_ARCHIVE_LAST_PROCESSED_OBS_ID` only exist on openmrs/openmrs-core#6171, which is still open, so none of them resolve against the published `openmrs-api:3.0.0-SNAPSHOT`. The build on this head fails with `cannot find symbol` on all five references (run 31025308449) and rerunning it won't help.

Could you note that dependency in the description and hold this until #6171 is merged, then rebase so CI actually gets to exercise the change?

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
```diff
@@ -378,10 +382,23 @@ public static void setGroupMembers(Obs obsGroup, Set<Obs> members) {
...
+		// Also include any archived children from obs_archive
+		ObsArchiveHelper archiveHelper = getObsArchiveHelperSafely();
+		if (archiveHelper != null && obs.getObsId() != null) {
+			List<Obs> archivedChildren = archiveHelper.getArchivedChildObs(obs.getObsId());
+			if (archivedChildren != null) {
+				members.addAll(archivedChildren);
+			}
+		}
+
+		return members.isEmpty() ? null : members;
```
dkayiwa, line 401: Every row in `obs_archive` is voided: the archiving task selects `FROM Obs o WHERE o.voided = :voided AND o.dateVoided < :cutoffDate` with `voided` bound to true. `Obs.getGroupMembers()` deliberately strips voided members (the no-arg form delegates to `getGroupMembers(false)`, which does `nonVoided.removeIf(BaseOpenmrsData::getVoided)`). So this unions voided obs into a property that has never carried them, on every request, including ones that never asked for deleted data. `GET /ws/rest/v1/obs/{uuid}` and `GET /ws/rest/v1/encounter/{uuid}?v=full` both serialize obs in the default representation, and that representation includes `groupMembers`, so with no `includeAll` anywhere in the request a client starts getting archived children nested under live parents. Any client written against the current contract, where `groupMembers` is by definition non-voided, has no reason to check `voided` on each member and will show deleted observations as current. Nothing throws, so neither the tests nor the logs will say a word.

The global property check in `getObsArchiveHelperSafely()` doesn't hold it back either. `obs.archive.last_processed_obs_id` is registered in `OpenmrsConstants.CORE_GLOBAL_PROPERTIES()` with a default of `-1` and inserted at startup by `Context.checkCoreDataset()`, so it is always present and never empty and the branch always passes. That leaves one global property lookup and one `obs_archive` query for every obs the API serializes, so an encounter with 40 obs adds 40 archive queries to a single fetch, on installs that have never turned archiving on. (`Encounter.getAllObsIncludingArchived()` carries a copy of the same check, worth fixing on #6171 while it is still open. `obs.archive.enabled` is the property that actually means what this is trying to test for.)

I'd drop this part entirely. The `doSearch` change covers RESTWS-1039 on its own: `Encounter.addObs` propagates the encounter down the whole group tree, so archived children keep their `encounter_id` and `getArchivedObsByEncounterId` already returns them in the flat result for `?encounter=X&includeAll=true`. And since `groupMembers` never showed voided children, there is no parity to restore here. If nesting archived children under their parent is genuinely wanted, it has to key off `includeAll`, and a static `@PropertyGetter` has no way to reach the `RequestContext`, so that belongs in its own change. Reverting also drops the module's dependency on `org.openmrs.api.impl.ObsArchiveHelper`, which is a core-internal class, and takes the three stray imports at the top of the file with it, including the second copy of `APIException`.

```suggestion
		if (obs.getGroupMembers() != null && !obs.getGroupMembers().isEmpty()) {
			return obs.getGroupMembers();
		}
		return null;
```

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ObsResource1_8.java
```diff
@@ -574,7 +591,12 @@ protected PageableResult doSearch(RequestContext context) {
 			if (enc == null)
 				return new EmptySearchResult();
 			
-			List<Obs> obs = new ArrayList<Obs>(enc.getAllObs(context.getIncludeAll()));
+			List<Obs> obs;
+			if (context.getIncludeAll()) {
```
dkayiwa, line 595: Nothing tests that an archived obs actually comes back, which is the whole of RESTWS-1039. `searchByEncounter_shouldGetVoidedObsIfIncludeAllIsTrue` in `ObsController1_9Test` covers the voided half of this branch, so the refactor itself is guarded, but nothing would notice if `getAllObsIncludingArchived()` stopped returning archived rows.

It is testable here: context sensitive tests build the schema with `hbm2ddl=create-drop`, and `ObsArchive` is `@Entity @Table(name = "obs_archive")` on the core branch, so the table exists in the test database and a dataset can seed a row. Please extend that test, or add one beside it, that seeds an archived obs for encounter `62967e68-96bb-11e0-8d6b-9b9415a91465` and asserts it comes back with `includeAll=true` and not without.

PR #755, opened 2026-08-19
RESTWS-1040: Add TypedSimpleObject<T> class

sohamdhande:

Adds `TypedSimpleObject<T>` (extends `SimpleObject`) as the foundation for RESTWS-1040, and migrates a small proof-of-concept batch of 5 resource classes from `new SimpleObject()` to `new TypedSimpleObject<>()` at their construction sites, ahead of a larger rollout across the rest of the codebase.

`TypedSimpleObject<T>` extends `SimpleObject` (not `LinkedHashMap` directly) per discussion with @chibongho — this preserves backward compatibility for modules relying on `SimpleObject` while migrating to REST 4.x. The `<T>` generic is currently unused (`TypedSimpleObject<?>`) as an interim step; follow-up tickets will fill in concrete types.

Migrated classes: `SystemInformationResource1_8`, `ServerLogResource1_8`, `ServerLogResource2_4`, `OrderEntryConfigResource1_10`, `ObsResource1_8`.

Method return type signatures were intentionally left as `SimpleObject` (constrained by the `Listable` interface contract) — since `TypedSimpleObject` extends `SimpleObject`, assignment compatibility is preserved and callers are unaffected. Only local variable declarations and construction sites were changed.

This is a small PoC batch to validate the migration pattern before scaling to the remaining resource classes.

## Issue I worked on

see https://openmrs.atlassian.net/browse/RESTWS-1040

- [x] I have **added tests** to cover my changes. (If you refactored existing code that was well tested you do not have to add tests) — added `TypedSimpleObjectTest` covering basic construction/put/get behavior; existing tests for migrated resources verified unchanged.
- [x] My pull request is **based on the latest changes** of the master branch. — *Note: based on 3.x branch, not master, per agreement with maintainer.*

### DIFF
--- a/omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
+++ b/omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
@@ -1,0 +1,104 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest;
+
+/**
+ * A type-parameterized extension of {@link SimpleObject} introduced as part of
+ * <a href="https://issues.openmrs.org/browse/RESTWS-1040">RESTWS-1040</a> /
+ * <a href="https://issues.openmrs.org/browse/RESTWS-1041">RESTWS-1041</a> to
+ * eventually carry accurate OpenAPI / Swagger type information for REST resource
+ * representations.
+ *
+ * <h3>Why this class exists</h3>
+ * <p>
+ * {@code SimpleObject} (which extends {@code LinkedHashMap<String, Object>}) is the
+ * universal return type for all REST resource handlers today. This works at runtime
+ * but prevents tooling (OpenAPI generators, IDE inspections, static analysis) from
+ * knowing <em>which</em> domain type a given map actually represents.
+ * </p>
+ * <p>
+ * {@code TypedSimpleObject<T>} adds a generic type parameter {@code <T>} so that
+ * resource methods can declare, for example,
+ * {@code TypedSimpleObject<Patient>} instead of plain {@code SimpleObject}.
+ * The parameter is erased at runtime and has <strong>no effect on serialization or
+ * map behaviour</strong> — it exists solely to improve compile-time documentation
+ * and to enable future OpenAPI schema generation.
+ * </p>
+ *
+ * <h3>Backward compatibility</h3>
+ * <p>
+ * This class intentionally extends {@link SimpleObject} (not
+ * {@code LinkedHashMap} directly) so that every {@code TypedSimpleObject} is
+ * assignment-compatible with {@code SimpleObject}. Modules that still consume
+ * the old {@code SimpleObject}-based API will continue to work unchanged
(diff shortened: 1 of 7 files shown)


omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
```diff
@@ -0,0 +1,104 @@
...
+ */
+public class TypedSimpleObject<T> extends SimpleObject {
+	
+	private static final long serialVersionUID = 1L;
+	
+	/**
+	 * Creates an empty {@code TypedSimpleObject}.
+	 */
+	public TypedSimpleObject() {
+		super();
```
chibongho: Nit: You probably copied this pattern from `SimpleObject`, but you can omit this line. A constructor automatically calls the empty super-contructor, unless you do `super(...)` yourself.

chibongho:
Looks great so far. Converting this as a draft for now as this will need to be done on the (upcoming) 4.x branch.

> Method return type signatures were intentionally left as SimpleObject (constrained by the Listable interface contract) — since TypedSimpleObject extends SimpleObject, assignment compatibility is preserved and callers are unaffected. Only local variable declarations and construction sites were changed.

I'd just modify this slightly. Yes, we should keep the `SimpleObject` return type for functions  in `Listable`, or any other interface / class meant to be extended, because changing it will force subclasses to update their return values (thus making this a huge breaking change). That said, for functions that are not meant to be overridden by classes in other modules, like `PageableResult.toSimpleObject()`, we should change their return type to `TypedSimpleObject`. (This does not need to be done as part of this PR, and does not change your approach.)

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_8/ServerLogResource1_8.java
```diff
@@ -39,7 +40,7 @@ public void setServerLogActionWrapper(ServerLogActionWrapper serverLogActionWrap
 	
 	@Override
 	public SimpleObject getAll(RequestContext context) throws ResponseException {
```
chibongho: We should be able to change the return type here to `TypedSimpleObject<?>`, as an overriding function is allowed to return a narrower type. While this doesn't affect how this function gets used, I'd like to get to a point where `SimpleObject` gets used as infrequently as possible, and we comment on why when we do use it instead of `TypedSimpleObject`.

sohamdhande:
Thanks for the review! Pushed a commit addressing both comments:

Removed the redundant super() call in TypedSimpleObject's no-arg constructor (copied the pattern from SimpleObject without thinking it through).
Narrowed ServerLogResource1_8#getAll to return TypedSimpleObject<?> via covariant return, since it's allowed here and doesn't affect callers. Also caught and removed the now-unused SimpleObject import in that file.

omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
```diff
@@ -0,0 +1,103 @@
...
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.webservices.rest;
+
+/**
```
chibongho, line 12: Let's clean up this comment, and make it more relevant to people using it. Can be as simple as this:

```
TypedSimpleObject<T> is a subclass of `SimpleObject` that allows us to store the type of its underlying data in the generic class `T`. This allows us to make better inference on the return type via reflection.
```

chibongho:
Looks good. Can you rebase on top the of 4.x branch, and make the PR merge onto that branch instead of master?

chibongho (approved):
LTGM.

PR #758, opened 2026-08-22
RESTWS-1048: Make PageableResult generic (PageableResult<T>)

sohamdhande:

Makes `PageableResult<T>` generic, per RESTWS-1048 (follow-up to RESTWS-1040/1041).

- `PageableResult.toSimpleObject` now returns `TypedSimpleObject<T>` and accepts `Converter<? super T>` (contravariant to support `DelegatingSubclassHandler`'s subclass-to-superclass-converter pattern).
- `BasePageableResult<T>` and `EmptySearchResult<T>` updated to match.
- `VisitResource1_9` updated — the only concrete resource forced to change, since it directly overrides `search()` rather than `doSearch()`/`doGetAll()`.
- `MainResourceController`/`MainSubResourceController` updated with minimal `@SuppressWarnings` for the existing dynamic-dispatch pattern (resources looked up by string name), per discussion with @chibongho — not restructuring `SearchHandler` to be generic.

**Open scope question for review:** Java allows raw-type overrides, so the ~42 concrete resource classes and 20 `SearchHandler` implementations currently compile unchanged and still return raw `PageableResult` under the hood — only `VisitResource1_9` was forced to update. This PR gets the core plumbing generic; propagating concrete types down to individual resources/search handlers is not yet done. Want to confirm this is the right scope for this ticket vs. a follow-up.

**Depends on:** #755 (`TypedSimpleObject<T>`) — cherry-picked that file onto this branch since #755 hasn't merged yet. Will rebase once #755 lands.

Build passes clean. Verified test-neutral vs master via clean rebuild on both branches (1793 tests run, 26 skipped, same 2 pre-existing failures on both — ruled out via side-by-side comparison, not just diff count).

## Issue I worked on

see https://openmrs.atlassian.net/browse/RESTWS-1048

- [x] My IDE is configured to follow the [code style](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.
- [x] I have added tests to cover my changes. — *No new behavioral logic to test beyond what `TypedSimpleObjectTest` (from #755) already covers; this PR is primarily type-signature propagation.*
- [x] All new and existing tests passed. — verified test-neutral vs master, see above.
- [x] My pull request is based on the latest changes of the master branch. — *based on 3.x/master per maintainer guidance; depends on #755 landing first.*


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/SearchHandler.java
```diff
@@ -29,5 +29,5 @@ public interface SearchHandler {
 	 * @return the result
 	 * @throws ResponseException
 	 */
-	PageableResult search(RequestContext context) throws ResponseException;
+	PageableResult<?> search(RequestContext context) throws ResponseException;
```
chibongho, line 32: Let's make `SearchHandler` class generic as well, even though we likely won't be using the generic class for generating the OpenAPI docs for `Resources` (in `MainResourceController` and `MainSubresourceController`). The `PageableResult` class is used by `Controllers` in other modules, and making this use the correct generic type for it is a good pattern to establish. Doesn't need to be in this PR.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingReadableResource.java
```diff
@@ -60,6 +61,7 @@ public Object retrieve(String uuid, RequestContext context) throws ResponseExcep
 	 * @see org.openmrs.module.webservices.rest.web.resource.api.Listable#getAll(org.openmrs.module.webservices.rest.web.RequestContext)
 	 */
 	@Override
+	@SuppressWarnings("unchecked")
 	public SimpleObject getAll(RequestContext context) throws ResponseException {
```
chibongho, line 65: This should return TypedSimpleObject

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
```diff
@@ -190,22 +192,23 @@ public void purge(String uuid, RequestContext context) throws ResponseException
...
+	protected PageableResult<T> doSearch(RequestContext context) {
+		return new EmptySearchResult<>();
 	}
 	
 	/**
 	 * @see org.openmrs.module.webservices.rest.web.resource.api.Listable#getAll(org.openmrs.module.webservices.rest.web.RequestContext)
 	 */
 	@Override
+	@SuppressWarnings("unchecked")
 	public SimpleObject getAll(RequestContext context) throws ResponseException {
```
chibongho, line 212: `TypedSimpleObject`

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingSubResource.java
```diff
@@ -172,7 +172,7 @@ public void purge(String parentUniqueId, String uuid, RequestContext context) th
 	@Override
 	public SimpleObject getAll(String parentUniqueId, RequestContext context) throws ResponseException {
```
chibongho, line 173: same here

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainSubResourceController.java
```diff
@@ -75,6 +75,7 @@ public Object retrieve(@PathVariable("resource") String resource, @PathVariable(
 	 * @return
 	 * @throws ResponseException
 	 */
+	@SuppressWarnings({ "rawtypes", "unchecked" })
 	@RequestMapping(value = "/{resource}/{parentUuid}/{subResource}", method = RequestMethod.GET)
 	@ResponseBody
 	public SimpleObject get(@PathVariable("resource") String resource, @PathVariable("parentUuid") String parentUuid,
```
chibongho, line 81: `TypedSimpleObject<?>`

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainResourceController.java
```diff
@@ -177,7 +177,7 @@ public Object purge(@PathVariable("resource") String resource, @PathVariable("uu
 	 * @return
 	 * @throws ResponseException
 	 */
-	@SuppressWarnings("rawtypes")
+	@SuppressWarnings({ "rawtypes", "unchecked" })
 	@RequestMapping(value = "/{resource}", method = RequestMethod.GET)
 	@ResponseBody
 	public SimpleObject get(@PathVariable("resource") String resource, HttpServletRequest request,
```
chibongho, line 183: same here

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/MainResourceController.java
```diff
@@ -219,7 +219,7 @@ public SimpleObject get(@PathVariable("resource") String resource, HttpServletRe
 	 * @return
 	 * @throws ResponseException
 	 */
-	@SuppressWarnings("rawtypes")
+	@SuppressWarnings({ "rawtypes", "unchecked" })
 	@RequestMapping(value = "/{resource}/search/{searchHandlerId}", method = {RequestMethod.GET, RequestMethod.POST})
 	@ResponseBody
 	public SimpleObject searchByHandler(@PathVariable("resource") String resource, @PathVariable("searchHandlerId") String searchHandlerId,
```
chibongho, line 225: same here

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/SubResourceSearchHandler.java
chibongho, line 15: This class is rarely used, but we should make it generic as well.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/PageableResult.java
```diff
@@ -9,14 +9,16 @@
...
 /**
  * Helper interface used by implementations that want to do pagination
+ *
+ * @param <T> the generic type of the list of results
  */
-public interface PageableResult {
+public interface PageableResult<T> {
 	
-	SimpleObject toSimpleObject(Converter<?> preferredConverter) throws ResponseException;
+	TypedSimpleObject<T> toSimpleObject(Converter<? super T> preferredConverter) throws ResponseException;
```
chibongho, line 22: The return type here is incorrect. I recommend creating a new Java type in `org.openmrs.module.webservices.rest.web` like this: 
```java
class PageableResultDto<T> {
  List<TypedSimpleObject<T>> results;
  Integer totalCount;
  List<HyperLink> links; 
}
```
Then return type of this function should be `TypedSimpleObject<PageableResultDto<T>>`.

This also means a lot of functions that return `TypedSimpleObject` also need fixing as well.

omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
```diff
@@ -38,6 +40,38 @@ public TypedSimpleObject(int initialCapacity) {
...
+	 * Adapts a {@link SimpleObject} to a {@link TypedSimpleObject}. If the object is
+	 * already a {@code TypedSimpleObject}, it is returned as-is. Otherwise, a new
+	 * {@code TypedSimpleObject} containing the same entries is created.
+	 *
+	 * @param <T> the target domain type
+	 * @param simpleObject the simple object to adapt
+	 * @return a {@code TypedSimpleObject} representing the same data, or null if input is null
+	 */
+	@SuppressWarnings("unchecked")
+	public static <T> TypedSimpleObject<T> from(SimpleObject simpleObject) {
```
chibongho, line 65: This feels wrong. It's fine as a utility function for tests if we move it to `TypedSimpleObjectTest.java`, but I don't think we should use it for real code. 

I think it does point to a need to also make `Searchable` and `Listable` generic as well. That part doesn't need to be part of this PR, but for this one let use `TypedSimpleObject<?>` (or even `SimpleObject`) instead of calling this function.

chibongho (changes requested):

sohamdhande:
Thanks for the detailed feedback, @chibongho! Addressed all three points in commit b10e8da6:

SubResourceSearchHandler now generic: SubResourceSearchHandler<T> extends SearchHandler<T>. Updated its only implementation, LocationAttributeSearchHandler1_9, to implements SubResourceSearchHandler<LocationAttribute>.

PageableResult.toSimpleObject introduced PageableResultDto<T> (with results: List<TypedSimpleObject<T>>, totalCount, links: List<Hyperlink>) as you suggested. PageableResult, BasePageableResult, and EmptySearchResult now construct and return TypedSimpleObject<PageableResultDto<T>> instead of the ad hoc map. Note: this generic typing terminates at the Searchable/Listable boundary those interfaces still return SimpleObject, per your note that generifying them is out of scope for this PR.

TypedSimpleObject.from() removed entirely from production code. Test coverage moved into TypedSimpleObjectTest.java. MainResourceController/MainSubResourceController reverted to plain SimpleObject return types, no more wrapper calls.

dkayiwa (commented):
Needs the XML output change flagged on `BasePageableResult` sorted out before merge. The rest of what I left is optional.

On your scope question: leaving the remaining raw `SearchHandler` implementations for a follow-up looks right to me. One thing worth knowing when that follow-up happens, generifying the implementations won't be enough on its own, because `RestService.getSearchHandler` and `getSearchHandlers` are themselves declared with the raw type, so the controllers keep seeing a raw `SearchHandler` (and keep needing the `unchecked` suppression) until those signatures move too.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BasePageableResult.java
```diff
@@ -48,18 +49,19 @@ public Long getTotalCount() {
...
-			Object converted = ConversionUtil.convertToRepresentation(match, context.getRepresentation(), preferredConverter);
+			Object converted = ConversionUtil.convertToRepresentation(match, context.getRepresentation(), (Converter) preferredConverter);
 			if (converted == ConversionUtil.PRIVILEGE_DENIED) {
 				continue;
 			}
 			results.add(converted);
 		}
 		
-		SimpleObject ret = new SimpleObject().add("results", results);
+		TypedSimpleObject<PageableResultDto<T>> ret = new TypedSimpleObject<PageableResultDto<T>>().add("results", results);
```
dkayiwa: Returning a `TypedSimpleObject` here renames the XML root element on every list and search response in the API, and nothing in the suite notices. `@XStreamAlias("object")` is declared on `SimpleObject`, and Java annotations aren't inherited, so XStream's mapper has no alias for the subclass and falls back to the fully qualified class name.

Same request on each branch (`GET /v1/concept` with `Accept: application/xml`, driven through `MainResourceController` and the real `MarshallingHttpMessageConverter`):

```
4.x:  <object><results><concept><uuid>...
758:  <org.openmrs.module.webservices.rest.TypedSimpleObject><results><concept><uuid>...
```

The search path does the same thing. If this merges as-is, any XML client anchored on the document root breaks on every paginated endpoint, silently and with a green build: the module's XML tests are all single-object retrieves, and the two that assert anything use `//uuid` and `//name/name`, which match at any depth, so none of them sees the root change.

Putting the alias on the subclass restores it. I tried that and `GET /v1/concept` goes back to `<object>` with `shouldGetAConceptByUuidInXML` still passing:

```java
@XStreamAlias("object")
public class TypedSimpleObject<T> extends SimpleObject {
```

That belongs in `TypedSimpleObject` rather than here, because the resources #755 already switched over have the same problem on 4.x today (`GET /v1/systeminformation` on the base branch already answers with `<org.openmrs.module.webservices.rest.TypedSimpleObject>`). A list-response XML test alongside it would close the gap that let this through.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingSubclassHandler.java
```diff
@@ -49,6 +49,6 @@ public interface DelegatingSubclassHandler<Superclass, Subclass extends Supercla
 	 * @throws ResourceDoesNotSupportOperationException if this resource does not support the
 	 *             operation
 	 */
-	PageableResult getAllByType(RequestContext context) throws ResourceDoesNotSupportOperationException;
+	PageableResult<?> getAllByType(RequestContext context) throws ResourceDoesNotSupportOperationException;
```
dkayiwa: `PageableResult<Subclass>` would serve better here than `PageableResult<?>`. The wildcard throws away the one piece of type information the callers have, which is why both `getAll` implementations now need `(Converter) this` plus a method-wide `@SuppressWarnings("unchecked")`. It also cancels out the `Converter<? super T>` contravariance you added to `toSimpleObject`, which exists for exactly this subclass-handler case.

Both call sites already declare `DelegatingSubclassHandler<T, ? extends T> handler`, so with `Subclass` in the signature `Converter<T>` satisfies `Converter<? super capture-of-? extends T>` and the call type-checks unaided:

```java
PageableResult<? extends T> result = handler.getAllByType(context);
return result.toSimpleObject(this);
```

I ran it: with that change plus both casts and both `@SuppressWarnings` deleted from `DelegatingCrudResource.getAll` and `BaseDelegatingReadableResource.getAll`, `mvn package` is green across the reactor. The three handlers in the tree (`DrugOrderSubclassHandler1_8`, `DrugOrderSubclassHandler1_10`, `TestOrderSubclassHandler1_10`) still override with a raw `PageableResult` and keep compiling.

The reason I'd bother: those two suppressions cover the whole `getAll` body, so an unsafe cast added in there later goes unreported.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BasePageableResult.java
```diff
@@ -48,18 +49,19 @@ public Long getTotalCount() {
 	 * <strong>Should</strong> not add property totalCount if context contains parameter totalCount which is false
 	 * <strong>Should</strong> not add property totalCount if context does not contains parameter totalCount
 	 */
+	@SuppressWarnings({ "unchecked", "rawtypes" })
```
dkayiwa: These two suppressions, and the `(Converter)` cast on line 57, look unnecessary. `ConversionUtil.convertToRepresentation` already declares its third parameter as a raw `Converter`, so a `Converter<? super T>` goes straight in. I compiled the method with the annotation and the cast both removed under `-Xlint:unchecked,rawtypes` and got zero warnings, with a deliberately raw `List` dropped into the same method as a control to confirm warnings would have shown up if there were any.

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java
```diff
@@ -25,6 +25,7 @@
 import org.openmrs.api.context.Context;
 import org.openmrs.customdatatype.CustomDatatypeUtil;
 import org.openmrs.module.webservices.rest.SimpleObject;
+import org.openmrs.module.webservices.rest.TypedSimpleObject;
```
dkayiwa, line 28: This import isn't used, and it's the entire diff for this file, so `VisitResource1_9` can come out of the PR altogether. The same leftover from the `TypedSimpleObject.from()` approach that b10e8da6 reverted sits in five other files: `MainResourceController` line 13, `MainSubResourceController` line 16, `DelegatingSubResource` line 19, `DelegatingCrudResource` line 19, `BaseDelegatingReadableResource` line 13.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/PageableResultDto.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Data transfer object representing a paginated list of results.
+ * 
+ * @param <T> the underlying type of the elements in the result set
+ */
+public class PageableResultDto<T> {
+	
+	private List<TypedSimpleObject<T>> results;
+	
+	private Integer totalCount;
```
dkayiwa: This wants to be `Long`. `BasePageableResult.getTotalCount()` returns `Long`, and that is the value written into the map, so the DTO whose whole job is to state the response shape currently disagrees with what the code puts there. `NeedsPaging` returns `(long) size` and `AlreadyPaged` stores a `Long` as well. The accessors on lines 38 and 42 need the same change.

```suggestion
	private Long totalCount;
```

omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
dkayiwa: The constructor `from()` was calling is still there, and nothing in production reaches it any more: `new TypedSimpleObject<>(map)` at TypedSimpleObject:48 has `mapConstructor_shouldPopulateEntries` as its only caller. That leaves the same shape you objected to, a production API kept alive by a test. Dropping both, along with the stray blank lines the removal left at TypedSimpleObject:54-56 and TypedSimpleObjectTest:121-122, would finish it off.

sohamdhande:
Thanks for the review, @dkayiwa especially catching the XML root regression, that's a nasty one to have slipped through silently. Addressed all five in commit 62106853.

XML root element (blocking) Added @XStreamAlias("object") to TypedSimpleObject rather than touching BasePageableResult, as you suggested, since this also repairs the same issue already present on 4.x via #755. Added shouldGetConceptsInXMLWithObjectRoot, asserting name(/*) equals object on a list response. Verified it's a meaningful test: it fails with the exact <org.openmrs.module.webservices.rest.TypedSimpleObject> root you described when the alias is removed, and passes once restored.

PageableResultDto.totalCount → Long done, field and accessors updated to match getTotalCount()/NeedsPaging/AlreadyPaged.

DelegatingSubclassHandler.getAllByType → PageableResult<Subclass> done, using your suggested PageableResult<? extends T> pattern. Removed the (Converter) casts and method-wide @SuppressWarnings("unchecked") from both getAll implementations; both compiled unaided as you predicted. Confirmed the three existing raw-type handlers (DrugOrderSubclassHandler1_8, DrugOrderSubclassHandler1_10, TestOrderSubclassHandler1_10) still compile.

BasePageableResult suppressions removed the @SuppressWarnings({"unchecked","rawtypes"}) and the cast on line 57. Confirmed zero warnings under -Xlint:unchecked,rawtypes (verified the lint flag was actually reaching javac via a deliberate raw-type control that did produce warnings, then removed it).

Unused TypedSimpleObject imports VisitResource1_9 reverted completely and drops out of the diff. Stripped the leftover import from the other five files.

Full reactor build green, 1803 tests / 0 failures / 14 skipped. Noted your point on RestService.getSearchHandler/getSearchHandlers for the SearchHandler follow-up will flag that in the follow-up ticket when we get there. Let me know if anything else needs adjusting.

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs1_9/VisitResource1_9.java
dkayiwa: Still here on 62106853: line 28 is the whole change to this file, and nothing else in it references `TypedSimpleObject`, so `VisitResource1_9` can come out of the PR entirely. Doesn't hold anything up, it's just a file in the diff with no reason to be there.

omod-common/src/main/java/org/openmrs/module/webservices/rest/TypedSimpleObject.java
dkayiwa: `TypedSimpleObject(Map<String, ?>)` is added by this PR rather than inherited from 4.x, and at 62106853 its only caller is still `mapConstructor_shouldPopulateEntries`. So it ships as public surface on a public class with nothing in production behind it. Deleting the constructor and that test, plus the extra blank lines at TypedSimpleObject:57-59 and TypedSimpleObjectTest:121-122, is the whole cleanup. Not a blocker.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingReadableResource.java
chibongho: Shouldn't this be `TypedSimpleObject<PageableResultDto<? extends T>>`?

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
```diff
@@ -191,15 +192,15 @@ public void purge(String uuid, RequestContext context) throws ResponseException
 	 */
 	@Override
 	public SimpleObject search(RequestContext context) throws ResponseException {
```
chibongho: `TypedSimpleObject<PageableResultDto<T>>`

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingCrudResource.java
chibongho: `TypedSimpleObject<PageableResultDto<? extends T>>`

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BasePageableResult.java
chibongho: We should be creating a `PageableResultDto` object here, and then converting it `TypedSimpleObject<PageableResultDto<T>>`. Doing `.add()` in various places leads to inaccurate typing, and  defeats the purpose of `TypedSimpleObject`.

We should have some sort of util function to convert a POJO of type `T` into `TypedSiimpleObject<T>`. See if there's anything re-useable in `ConversionUtil` to help with this, and create one there if not.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/BaseDelegatingReadableResource.java
```diff
@@ -86,23 +86,23 @@ public SimpleObject getAll(RequestContext context) throws ResponseException {
...
 	 */
-	public PageableResult doGetAll(RequestContext context) {
+	public PageableResult<T> doGetAll(RequestContext context) {
 		throw new ResourceDoesNotSupportOperationException();
 	}
 	
 	/**
 	 * @see org.openmrs.module.webservices.rest.web.resource.api.Searchable#search(org.openmrs.module.webservices.rest.web.RequestContext)
 	 */
 	public SimpleObject search(RequestContext context) throws ResponseException {
```
chibongho: `TypedSimpleObject`

PR #759, opened 2026-08-25
RESTWS-1049: Fix broken tests in master branch 

chibongho:

Upstream core changes broke some of the tests in the master branch for the REST module here. See failing tests [here](https://github.com/openmrs/openmrs-module-webservices.rest/actions/runs/32291007969/job/96191710037). See summary by Claude below. Note that I plan to either remove `swagger-core:1.6.2` or upgrade it to a significantly more recent version soon, so the fix to problem 1 will likely change later. 

Also, if `master` branch in `openmrs-core` is stable enough, we should consider adding a Github action there to build against this module.

Summary by Claude:

**Problem 1 (guava).** openmrs-core `c4aca3af4` (#6233) added `OutboxEventRegistry`, a Spring bean
whose constructor calls `CacheBuilder.expireAfterWrite(Duration.ofMinutes(10))` — an overload that
exists only in the **jre** flavor of Guava, which core manages. But `omod-common` was resolving
`guava:27.0.1-android` via `io.swagger:swagger-core:1.6.2`, which wins Maven's nearest-declaration
mediation. Since the bean loads with every Spring context, all 112 context-sensitive tests in the
module failed at once with `NoSuchMethodError`. Latent defect here, not an upstream regression.
**Fix:** pin `guava:33.7.0-jre` in the root pom's `dependencyManagement`; packaging unchanged.

**Problem 2 (fixtures).** openmrs-core `1b5421a9a` (TRUNK-6712, #6383) added `RolePrivilegeCache` and
rewrote `UserContext.hasPrivilege` — role privileges now come from a cache populated by a daemon
thread that reloads the role from the DB and fails closed if absent, where the old path walked the
in-memory role graph. Both fixtures depended on that removed behaviour: one attached a role purely in
memory, the other saved one inside the test's rolled-back transaction. **Fix:** core's own adaptation
pattern from that commit — `Context.addProxyPrivilege(...)` in try/finally. Test-only.

## Issue I worked on

see https://openmrs.atlassian.net/browse/RESTWS-1049

  No? -> execute above command

### DIFF
--- a/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
+++ b/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
@@ -195,7 +195,8 @@ public class ConversionUtil1_9Test extends BaseModuleWebContextSensitiveTest {
                 "User without 'Get Patients' privilege should NOT see patient data through custom representation");
 
         // Positive control: the same user and the same representation, but now holding patient access.
-        // Without this, the assertion above would also pass if the privilege check never ran at all.
+        // Without this, the assertion above would pass even if the patient were unreachable on this
+        // visit for a reason unrelated to privileges, which would make it vacuous.
         Object allowedResult;
         try {
             Context.addProxyPrivilege("Get Visits");
--- a/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
+++ b/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
@@ -231,26 +231,10 @@ public class ConversionUtil1_9Test extends BaseModuleWebContextSensitiveTest {
         person.addName(new PersonName("Limited", null, "User"));
         Context.getPersonService().savePerson(person);
 
-        Role role = new Role("Limited Role");
-        role.setDescription("Role with limited privileges for testing");
-        for (String privName : new String[] { "Get Visits" }) {
-            Privilege priv = userService.getPrivilege(privName);
-            if (priv == null) {
-                priv = new Privilege(privName);
-                priv.setDescription(privName);
-                userService.savePrivilege(priv);
-            }
-            role.addPrivilege(priv);
-        }
-        userService.saveRole(role);
-
         User user = new User(person);
         user.setUsername("limited_user");
-        user.addRole(role);
         for (Role r : new ArrayList<>(user.getAllRoles())) {
-            if (!r.getRole().equals("Limited Role")) {
-                user.removeRole(r);
-            }
+            user.removeRole(r);
         }
         userService.createUser(user, "LimitedTest123");
     }
--- a/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
+++ b/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
@@ -17,8 +17,6 @@ import org.openmrs.Encounter;
 import org.openmrs.Obs;
 import org.openmrs.Person;
 import org.openmrs.PersonName;
-import org.openmrs.Privilege;
-import org.openmrs.Role;
 import org.openmrs.User;
 import org.openmrs.Visit;
 import org.openmrs.api.ConceptService;
@@ -32,7 +30,6 @@ import org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs1_8.Encounte
 import org.openmrs.web.test.jupiter.BaseModuleWebContextSensitiveTest;
 import org.springframework.beans.factory.annotation.Autowired;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import static org.hamcrest.core.Is.is;
--- a/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
+++ b/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
@@ -233,9 +233,6 @@ public class ConversionUtil1_9Test extends BaseModuleWebContextSensitiveTest {
 
         User user = new User(person);
         user.setUsername("limited_user");
-        for (Role r : new ArrayList<>(user.getAllRoles())) {
-            user.removeRole(r);
-        }
         userService.createUser(user, "LimitedTest123");
     }
 }
--- a/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
+++ b/omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
@@ -166,10 +166,10 @@ public class ConversionUtil1_9Test extends BaseModuleWebContextSensitiveTest {
         Context.logout();
         Context.authenticate("limited_user", "LimitedTest123");
         assertTrue(Context.isAuthenticated());
-        // "Limited Role" grants nothing here: role privileges resolve by name from the database, and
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -152,6 +152,13 @@
 				<version>${jacksonVersion}</version>
 				<scope>provided</scope>
 			</dependency>
+			<!-- swagger-core pulls in the "android" flavor of Guava, which lacks the java.time
+			     overloads that openmrs-api relies on; pin the "jre" flavor used by openmrs-core -->
+			<dependency>
+				<groupId>com.google.guava</groupId>
+				<artifactId>guava</artifactId>
+				<version>33.7.0-jre</version>
```
dkayiwa: `33.7.0-jre` is a patch behind what the platform actually ships. Core's BOM moved to `33.7.1-jre` in openmrs/openmrs-core#6485 on 22 Aug, and `openmrs-api` declares guava with no version of its own, so this pin now overrides core rather than following it. Worth bumping, though nothing breaks if you'd rather not.

```suggestion
				<version>33.7.1-jre</version>
```

It shows up outside `omod-common` as well. In `omod`, `openmrs-api -> guava` sits at depth 2 and already beats `omod-common -> swagger-core -> guava` at depth 3, so master's packaged omod contains `lib/guava-33.7.1-jre.jar`; with the pin it contains `lib/guava-33.7.0-jre.jar`, and that one jar was the only difference between the two omods I built. So this fixes `omod-common` and at the same time moves `omod` a patch behind the platform it runs in. The standing cost is that the literal has to be hand-bumped whenever core's moves, and whoever hits the next `NoSuchMethodError` after a core guava bump has to work out that this line, not core, is holding the old version.

If you'd rather not carry a literal at all, excluding guava from `swagger-core` in `omod-common/pom.xml` fixes the same thing: I ran master with only that exclusion and omod-common's 104 tests all pass, leaving just the two fixture failures this PR also fixes. It is the bigger change, because guava then arrives at `provided` scope from openmrs-api and six jars leave the omod's `lib/` (guava, failureaccess, listenablefuture, error_prone_annotations, j2objc-annotations, jspecify). Given you plan to drop swagger-core soon, bumping the literal is probably the right size for now.

omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
```diff
@@ -161,12 +161,15 @@ public void convertToRepresentation_shouldNotLeakPatientDataThroughVisitCustomRe
...
+        // Create and authenticate as a user without "Get Patients"/"Get People"
         createLimitedUser();
         Context.logout();
         Context.authenticate("limited_user", "LimitedTest123");
         assertTrue(Context.isAuthenticated());
-        assertTrue(Context.hasPrivilege("Get Visits"));
+        // "Limited Role" grants nothing here: role privileges resolve by name from the database, and
+        // this role is created inside the test's uncommitted transaction. The visit is fetched and
+        // converted under a proxy privilege below instead. What matters here is the absence of
+        // patient access, which is what this test is about.
```
dkayiwa: Since the role genuinely grants nothing now, the role and privilege construction in `createLimitedUser` is inert, and it is the first thing a reader will take for the mechanism under test: a role called "Limited Role" holding "Get Visits". I deleted that whole `new Role(...)` / `savePrivilege` / `saveRole` / `addRole` block locally and all three tests still pass, so it can go, which would leave `createLimitedUser` doing what its name now means, creating a user with no privileges. Your call whether to trim it or keep it and say in the helper that it is deliberately dead.

chibongho: Removed adding of "Limited Role" in the `createLimitedUser()` function

omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
```diff
@@ -203,26 +231,10 @@ private void createLimitedUser() {
...
         User user = new User(person);
         user.setUsername("limited_user");
-        user.addRole(role);
         for (Role r : new ArrayList<>(user.getAllRoles())) {
-            if (!r.getRole().equals("Limited Role")) {
-                user.removeRole(r);
-            }
+            user.removeRole(r);
         }
         userService.createUser(user, "LimitedTest123");
```
dkayiwa: With the role gone this loop cannot do anything: `new User(person)` leaves `roles` null, and `User.getAllRoles()` returns an empty set when `getRoles()` is null, so the body never runs. Dropping it frees the last uses of the `Role` and `ArrayList` imports, and `Privilege` has been unused since the last commit too. I removed the loop and those three imports locally and all three tests in the class still pass. Nothing breaks if you leave it, though it is the last piece of the role machinery still reading as setup.

```suggestion
        userService.createUser(user, "LimitedTest123");
```

omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
dkayiwa: The comment at line 169 still explains why "Limited Role" grants nothing, but the role is gone now, so its only remaining mention in the file is the explanation of a fixture that no longer exists. Worth keeping the reason, since it is why the proxies are there, just not the name. Something like:

```java
        // The user holds no roles at all: role privileges resolve by name from the database, so a role
        // created inside the test's uncommitted transaction would grant nothing anyway. The visit is
        // fetched and converted under a proxy privilege below instead. What matters here is the
        // absence of patient access, which is what this test is about.
```

omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
```diff
@@ -179,13 +182,38 @@ public void convertToRepresentation_shouldNotLeakPatientDataThroughVisitCustomRe
...
+        } finally {
+            Context.removeProxyPrivilege("Get Visits");
+        }
         assertInstanceOf(SimpleObject.class, limitedResult);
         SimpleObject limitedVisit = (SimpleObject) limitedResult;
         assertFalse(limitedVisit.containsKey("patient"),
                 "User without 'Get Patients' privilege should NOT see patient data through custom representation");
+
+        // Positive control: the same user and the same representation, but now holding patient access.
+        // Without this, the assertion above would also pass if the privilege check never ran at all.
```
dkayiwa: This has it the wrong way round: if the privilege check never ran, the assertion above would fail, not pass. With no gate the patient property comes through and `assertFalse(limitedVisit.containsKey("patient"))` trips. I checked by making the `getRequiredGetPrivilege` branch in `ConversionUtil` unreachable and rerunning the class, and the failure landed on line 197, not down here.

The block is still worth having, just for the other reason: it rules out `visitAsLimited` never yielding a patient at all, say because the association cannot be loaded under the limited user's session, which would let the negative assertion pass with the gate working perfectly well. That is what the message on line 216 already says.

```suggestion
        // Without this, the assertion above would pass even if the patient were unreachable on this
        // visit for a reason unrelated to privileges, which would make it vacuous.
```

omod/src/test/java/org/openmrs/module/webservices/rest/web/ConversionUtil1_9Test.java
dkayiwa: The three imports the loop was the last user of are still there: `org.openmrs.Role` on line 21, `java.util.ArrayList` on line 35, and `org.openmrs.Privilege` on line 20, which went unused when the role block came out. Each is now referenced only by its own import statement. I dropped all three locally and the class compiles with all three tests green.

dkayiwa (approved):

PR #754, opened 2026-08-17
RESTWS-1043: Deprecate Swagger model getters in Resource classes

sohamdhande:
## Description of what I changed

Marks `getGETModel(rep)`, `getCREATEModel(rep)`, and `getUPDATEModel(rep)` as `@Deprecated` across all Resource classes that implement them, per RESTWS-1043.

These functions currently duplicate (and sometimes contradict) the property-declaring functions (`getRepresentationDescription`, `getCreatableProperties`, `getUpdatableProperties`), which should be the source of truth for Swagger/OpenAPI typings going forward. Since the Swagger JSON has been broken since 3.7 without much attention, it's safe to assume no one depends on the current output.

This is Phase 1: deprecation only, no removal. Two internal callers still reference these methods and will need updates before removal in Phase 2 (tracked separately, blocked on the 4.x branch): `SwaggerSpecificationCreator.java` and `ServerLogActionWrapper2_4.java`.

`@deprecated` version is set to `3.6.0` since this change applies to the 3.x branch.

## Issue I worked on

see https://openmrs.atlassian.net/browse/RESTWS-1043

## Checklist: I completed these to help reviewers :)

- [x] My IDE is configured to follow the [code style](https://wiki.openmrs.org/display/docs/Java+Conventions) of this project.
- [x] I have added tests to cover my changes. (If you refactored existing code that was well tested you do not have to add tests) — *N/A: this is a Javadoc/annotation-only change (adding `@Deprecated`), no behavioral change to test.*
- [x] I ran `mvn clean package` right before creating this pull request and added all formatting changes to my commit.
- [x] All new and existing tests passed.
- [x] My pull request is based on the latest changes of the master branch. — *Note: based on 3.x branch, not master, as agreed with maintainer.*

PR #760, opened 2026-09-02
RESTWS-1043: Remove Swagger

chibongho:

This PR removes (a quite old) Swagger dependency in the 4.x branch. I'll add back a new version as dependency at some point, but mostly for adding type hints with annotations, and not doc generation.

- Resource classes' `getGETModel()`, `getCREATEModel()` and `getUPDATEModel()` functions, which are deprecated in the 3.x branch in https://github.com/openmrs/openmrs-module-webservices.rest/pull/754, are removed.
- `SwaggerSpecificationController`, which serves the Swagger JSON, and associated data classes are removed
- Delete test classes that were made purely for Swagger.
- The `SWAGGER_QUIET_DOCS_GLOBAL_PROPERTY_NAME` GP is now un-used and removed
- I kept `SwaggerDocController`, which serves the Swagger UI endpoint at `/module/webservices/rest/apiDocs`. The UI is broken already as `SwaggerSpecificationController` hasn't worked in a while, so this change is not making it worse. I'm keeping that route with the expectation that we'll have it serve OpenAPI docs that at generated at build time and bundled into module JAR files. 

Note: At some point I'll need to forward-port this to the `master` branch (on `5.0.0-SNAPSHOT`)

## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-1043

  No? -> execute above command


omod-common/pom.xml
```diff
@@ -29,26 +29,6 @@
 			<artifactId>evo-inflector</artifactId>
 			<version>1.2.2</version>
 		</dependency>
-
-		<dependency>
-			<groupId>io.swagger</groupId>
-			<artifactId>swagger-core</artifactId>
```
dkayiwa, line 35: Taking this out is what turns the Java 21 job red. `ClearDbCacheController2_0Test.clearDbCache_shouldEvictTheEntityFromTheCaches` fails at line 64, `assertTrue(sessionFactory.getCache().containsEntity(PERSON_NAME_CLASS, ID_2))`, and that test isn't touched anywhere in this diff.

I isolated it locally on JDK 21, same machine, `mvn -pl omod test` each time:

| variant | tests | result |
| --- | --- | --- |
| `origin/4.x` | 1813 | pass |
| `origin/4.x`, minus exactly the five test files this PR deletes | 1801 | pass |
| this branch | 1801 | **fail**, ClearDbCacheController2_0Test:64 |
| this branch with only the `io.swagger:swagger-core` block put back in this pom, every code change kept | 1801 | pass (ran it twice) |

So it isn't the test classes you dropped, and it isn't the resource edits. It's the dependency leaving the classpath. Comparing `dependency:list` between the two branches, the only artifacts that actually disappear are swagger-core, swagger-models and swagger-annotations; guava, commons-lang3, jspecify, error_prone_annotations and j2objc-annotations just move from compile to provided, so tests still see them.

Nothing in production is broken by this. Line 64 is a precondition assertion that runs before the controller under test is even called, and the test is fragile to start with: run `ClearDbCacheController2_0Test` on its own against `origin/4.x`, with none of your changes, and a different method in it fails (`clearDbCache_shouldEvictAllEntitiesFromTheCaches:122`). It leans on second-level cache state it never sets up, and losing three jars off the classpath is enough to disturb whatever was making that true.

It still needs sorting before this merges. As it stands 4.x goes red on Java 21 and stays red, so the branch build stops being a signal and whoever opens the next PR inherits a failing baseline. The fix belongs in the test, having it establish the cache state it asserts on, either here or in a small PR landed ahead of this one.

omod-common/pom.xml
```diff
@@ -29,26 +29,6 @@
 			<artifactId>evo-inflector</artifactId>
```
dkayiwa: `evo-inflector` can come out with the rest of it. Its only caller was `SwaggerSpecificationCreator` (the plural-of-resource-name helper near the bottom of that class on 4.x), and on this branch `git grep -i inflector` matches nothing outside this dependency block. It is still being shipped: `lib/evo-inflector-1.2.2.jar` is in the .omod this branch builds.

Same story for the `jackson-dataformat-yaml` test dependency in `omod/pom.xml`. Its comment, "to support tests without bundling", was about swagger-core's YAML factory, and nothing on the branch touches YAML now. Raising it here because that file isn't in the diff, so there's nowhere to comment on it directly.

omod/src/main/java/org/openmrs/module/webservices/rest/web/controller/SwaggerSpecificationController.java
```diff
@@ -1,45 +0,0 @@
...
-import org.openmrs.module.webservices.docs.swagger.SwaggerSpecificationCreator;
-import org.springframework.stereotype.Controller;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
-import org.springframework.web.bind.annotation.ResponseBody;
-
-import javax.servlet.http.HttpServletRequest;
-
-@Controller("SwaggerSpecificationController")
-@RequestMapping("/module/webservices/rest/swagger.json")
```
dkayiwa, line 23: With this gone, `/module/webservices/rest/apiDocs.htm` can only ever render Swagger UI's "Failed to load API definition" box. apiDocs.jsp still points SwaggerUIBundle at `${pageContext.request.contextPath}/module/webservices/rest/swagger.json`, nothing serves that path any more, and the advanced administration screen still links to the page (AdminSection line 56). The README's "API Documentation" section still describes it as working, too.

I'm not disputing that the page was already broken, and holding the route for the generated OpenAPI docs makes sense to me. It's the pieces that make the breakage visible and expensive that I'd still take out. Removing the AdminSection entry is one line and stops admins being sent somewhere that can only error. And the vendored Swagger UI 3.0.10 under `omod/src/main/webapp/resources/js` is 1.77 MB on disk, 457 KB of the 1.99 MB .omod this branch produces, for assets that now have nothing to load.

My preference would be to drop the JSP, the admin link and the swagger-ui directory here, and bring a current Swagger UI back alongside the OpenAPI work, keeping `SwaggerDocController` as the placeholder so the URL stays reserved. If you'd rather leave them in place, the admin link is the one I'd still remove.

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_1/PersonNameResource2_1.java
```diff
@@ -9,7 +9,6 @@
  */
 package org.openmrs.module.webservices.rest.web.v1_0.resource.openmrs2_1;
 
-import io.swagger.models.Model;
 import org.openmrs.PersonName;
 import org.openmrs.module.webservices.rest.web.annotation.SubResource;
 import org.openmrs.module.webservices.rest.web.representation.Representation;
```
dkayiwa: Unused now, and so are 13 others the method removals orphaned. I stripped all 14 and the reactor still packages clean, so none of them are load-bearing:

- `BaseDelegatingSubclassHandler`: `Representation`
- `PersonResource1_11`: `DefaultRepresentation`, `FullRepresentation`
- `ConceptClassResource1_8`: `Representation`
- `ConceptNameResource1_8`: `ConceptNameType`
- `EncounterTypeResource1_8`: `Representation`
- `CustomDatatypeHandlerResource1_9`: `DefaultRepresentation`, `FullRepresentation`
- `EncounterRoleResource1_9`: `Representation`
- `DatabaseChangeResource2_0`: `liquibase.changelog.ChangeSet`
- `PersonNameResource2_1`: `Representation`
- `DiagnosisResource2_2`: `ConditionVerificationStatus`
- `DiagnosisResource2_5`: `DefaultRepresentation`, `FullRepresentation`

An optimize-imports pass over the touched files clears them. `AlertResource2_0`'s `PrivilegeConstants` and `MedicationDispenseResource2_6`'s `Order` are unused as well, but were already that way on 4.x, so leave those if you want to keep the diff to swagger.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingResourceHandler.java
```diff
@@ -89,39 +88,4 @@ public interface DelegatingResourceHandler<T> extends DelegatingPropertyAccessor
...
-	/**
-	 * Returns a {@link Model} object representing GET representation schema for the resource.
-	 * 
-	 * @param rep representation type under which the resource {@link Model} should be fetched. It
-	 *            can take {@link Representation#DEFAULT}, {@link Representation#REF}, or
-	 *            {@link Representation#FULL}
-	 * @return a {@link Model} object or null in case if such model does not exist or not
-	 *         documented.
-	 */
-	Model getGETModel(Representation rep);
```
dkayiwa, line 102: Worth spelling out in the 4.0.0 release notes: for modules already built against 3.x, this is a runtime break and not just a recompile.

Those modules declare `public Model getGETModel(Representation)` on their resource classes, and they take webservices.rest at `provided` scope (stockmanagement, idgen, queue and bedmanagement all do), so they never bundled swagger-core themselves, they relied on this module's `lib/` for it. `Class.getMethods()` throws `NoClassDefFoundError` when any public method's return type can't be resolved, and `ReflectionUtil.findPropertyGetterMethod` calls exactly that on the handler class every time `BaseDelegatingResource.getProperty` converts a property. I checked the JVM behaviour with a standalone repro (compile a class against a type, delete the type's class file, then reflect over the class): it loads and instantiates fine, then `getMethods()` throws, and so does `getMethod()` for an unrelated method on the same class.

Since `require_module` versions are minimums, an unrebuilt module still starts against 4.0.0 and then throws `NoClassDefFoundError: io/swagger/models/Model` out of REST property resolution, rather than failing with anything legible at startup. More than a dozen modules in my local checkouts carry these overrides. Nothing to change in the code here, it just seems worth saying out loud so people rebuild rather than debug it.

omod-common/pom.xml
chibongho: This is incorrect. I re-ran and the test passed. It's an intermittent error that we should diagnose and fix, but is outside the scope of this PR.

omod-common/pom.xml
chibongho: Removed both.

omod/src/main/java/org/openmrs/module/webservices/rest/web/controller/SwaggerSpecificationController.java
chibongho: Even the admin link should not be a concern as I don't expect any consumers of the package before this route gets fixed.

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_1/PersonNameResource2_1.java
chibongho: Removed unused imported.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/impl/DelegatingResourceHandler.java
chibongho: Not in the PR, but yes, we'll be calling this out as a breaking change of 4.0.0

omod-common/pom.xml
dkayiwa: `ClearDbCacheController2_0Test` holds up at the head. Four consecutive `mvn -pl omod test` runs on 30ad87ee under JDK 21, same machine as the runs above: 1801 tests, 0 failures each time, and the class itself 4 of 4 on every run. The PR's own Java 21 job is green too.

The isolation behaviour turns out to be the same on both refs. `mvn -pl omod test -Dtest=ClearDbCacheController2_0Test`, JDK 21:

| ref | result |
| --- | --- |
| `origin/4.x` | 4 run, 1 failure, `clearDbCache_shouldEvictAllEntitiesFromTheCaches:122` |
| `30ad87ee` | 4 run, 1 failure, same method, same line |

So the classpath explanation I gave doesn't hold, and neither does the blocker. The test does lean on second-level cache state it never establishes, but it does that identically on 4.x, so it is a pre-existing fragility to fix on its own terms rather than something this PR has to carry.

mseaton (approved):
Great to see this module simplified and on the way to much improved docs!

PR #764, opened 2026-09-09
RESTWS-1045: Create TypedSearchHandler; convert EncounterSearchHandle…

chibongho:

This PR introduces `TypedSearchHandler`, and migrates one search handler, `EncounterSearchHandler2_0`, to extend it.

Currently, a `SearchHandler` declares what inputs (required or optional) it expects via its `SearchConfig`. This means it's possible to enumerate its inputs via reflection already. although we have no information on their expected types. `TypedSearchHandler` allows the SearchHandler to define an `Params` type that that enumerates all of the its expected inputs while also typing them. Admittedly, the typing benefit is limited right now because each input is a query param, which means its possible type can only be `String` or `String[]` (via repeated query parmms like `?visit=a&visit=b`). However, as we migrate more `SearchHandlers` to `TypedSearchHandler`, we can add first class deserialization support to types like `Date`, `Number` or `Boolean`. (This will also make for better typings in the generated TypeScript clients.)

A secondary benefit is this also gets us closer to supporting sending params for search handler via POST request body, which allows us to work around the URL character limits we face with query params. (See [RESTWS-1050](https://openmrs.atlassian.net/browse/RESTWS-1050)). Most search handlers now do `context.getRequest().getParameter(..)` to read the their inputs, but that only reads query params. By lifting the param parsing into `TypedSearchHandler`, we have one central place to read the inputs, which should allow us to add support for also reading params from POST request body easier.

Note: this PR paints a `TypedSearchHandler` that's simpler than it should be, as 3 (out of ~20) SearchHandlers in this module will not be able to migrate to using it as is. Confusingly, a SearchHandler does not just expect one set of inputs, but can expect one of several sets of inputs; See [RelationshipSearchHandler1_8](https://github.com/openmrs/openmrs-module-webservices.rest/blob/e8838aae8cd701aa789b2eca8d6531e963bd8eef/omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs1_8/RelationshipSearchHandler1_8.java). That said, I have made a local PoC change that allows for support those SearchHandlers as well, and it's not significantly more complex.

Testing done:
- Existing tests pass.
- Manually invoked the EncounterSearchhandler with `/encounter/search/default?patient=<uuid>` to verify that it still works as expected.

## Issue I worked on

see https://issues.openmrs.org/browse/RESTWS-1045

  No? -> execute above command

[RESTWS-1050]: https://openmrs.atlassian.net/browse/RESTWS-1050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
```diff
@@ -0,0 +1,129 @@
...
+	public final PageableResult search(RequestContext context) throws ResponseException {
+		Class<P> paramsClass = searchConfig.getParamsClass();
+		HttpServletRequest request = context.getRequest();
+		ObjectMapper objectMapper = createObjectMapper();
+
+		// read from query params
+		try {
+			P params = objectMapper.convertValue(request.getParameterMap(), paramsClass);
+			assertRequiredParamsPresent(params);
+			return page(search(params), context);
```
chibongho, line 69: Almost all `SearchHandlers` wrap the returned search results in a `NeedPaging`, so I have it do that by default. I did find one SearchHandler that returns a `AlreadyPaged` object instead though. I'll make this behavior overridable in a subsequent PR.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
```diff
@@ -0,0 +1,129 @@
...
+	 * handles array-to-string coercion (for query parameter maps where values
+	 * are String arrays).
+	 */
+	private ObjectMapper createObjectMapper() {
+		ObjectMapper objectMapper = new ObjectMapper();
+		// Ignore query parameters that aren't fields on the params class - notably the framework's
+		// special request parameters (limit, startIndex, v, s, ...) which ride along on every request
+		// and are handled elsewhere. Missing required params are still caught by
+		// assertRequiredParamsPresent(); only unrecognized extras are silently dropped.
+		objectMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
chibongho, line 86: This keeps the existing behavior of silently allowing unexpected query params.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
```diff
@@ -0,0 +1,129 @@
...
+		ObjectMapper objectMapper = createObjectMapper();
+
+		// read from query params
+		try {
+			P params = objectMapper.convertValue(request.getParameterMap(), paramsClass);
+			assertRequiredParamsPresent(params);
+			return page(search(params), context);
+		} catch(IllegalArgumentException e) {
+			throw new IllegalRequestException("Unable to read search request parameters", e);
+		}
```
dkayiwa, line 72: The `try` wraps `search(params)` as well as the binding, so an `IllegalArgumentException` raised by a subclass's own search logic reaches the client as a 400 saying "Unable to read search request parameters". I checked with a handler whose `search(P)` throws `IllegalArgumentException`: the caller gets `IllegalRequestException: Unable to read search request parameters`, and the real message survives only in the cause.

Narrowing the try to the bind keeps that message truthful:

```suggestion
		P params;
		try {
			params = objectMapper.convertValue(request.getParameterMap(), paramsClass);
		} catch (IllegalArgumentException e) {
			throw new IllegalRequestException("Unable to read search request parameters", e);
		}
		assertRequiredParamsPresent(params);
		return page(search(params), context);
```

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
```diff
@@ -0,0 +1,129 @@
...
+	@Override
+	public final TypedSearchConfig<R, P> getSearchConfig() {
+		return searchConfig;
+	}
+
+	@Override
+	public final PageableResult search(RequestContext context) throws ResponseException {
+		Class<P> paramsClass = searchConfig.getParamsClass();
+		HttpServletRequest request = context.getRequest();
+		ObjectMapper objectMapper = createObjectMapper();
```
dkayiwa, line 63: The mapper is rebuilt on every request, and Jackson caches bean deserializers per mapper, so `Params` gets re-introspected on each call instead of once. I measured the exact `createObjectMapper()` plus `convertValue` from this file against an `EncounterSearchHandler2_0.Params`-shaped POJO (Jackson 2.19.1, 20k iterations after warmup): 27.1 µs/op building a mapper each call, 0.9 µs/op with one shared mapper.

26 µs is nothing next to the query this handler is about to run, so I wouldn't hold the PR for it, but an `ObjectMapper` is thread safe once configured and this one never varies per request, so `private static final ObjectMapper OBJECT_MAPPER = createObjectMapper();` picks it up for free.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
```diff
@@ -0,0 +1,129 @@
...
+			}
+		});
+		objectMapper.registerModule(module);
+		return objectMapper;
+	}
+
+	private void assertRequiredParamsPresent(P params) throws ResponseException {
+		try {
+			Class<P> paramsClass = searchConfig.getParamsClass();
+			for(Field f : paramsClass.getDeclaredFields()) {
```
dkayiwa, line 106: This loop and `TypedSearchConfig.toSearchQuery` walk the same fields but disagree about which of them are parameters: `toSearchQuery` skips synthetic and static fields, this one doesn't. So a params class that declares any static field ends up with a `SearchQuery` that correctly ignores it and a runtime check that treats it as required.

I tried adding `static String NOT_A_PARAMETER;` to a params class. The declared required parameters come out as `[patient]`, and `?patient=some-uuid` is then rejected with `InvalidSearchException: Missing required search parameter: NOT_A_PARAMETER`.

The `$jacocoData` field your comment in `toSearchQuery` mentions is real, incidentally. Under the coverage build `EncounterSearchHandler2_0.Params.getDeclaredFields()` does return it (static, synthetic, `boolean[]`) and this loop does treat it as required. It only gets through because JaCoCo's instrumented constructor initialises it before the check reads it.

Nothing breaks in the encounter handler as written, so this isn't blocking, but it is the class the other handlers get migrated onto, and I'd rather the "is this field a declared parameter" test lived in one place both loops call.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
```diff
@@ -0,0 +1,129 @@
...
+		}
+	}
+
+	private PageableResult page(List<D> results, RequestContext context) {
+		if(results.isEmpty()) {
+			return new EmptySearchResult();
+		} else {
+			return new NeedsPaging<D>(results, context);
+		}
+	}
```
dkayiwa, line 126: Short-circuiting an empty result set to `EmptySearchResult` drops `totalCount`, and I think that needs fixing before this merges. `EmptySearchResult.toSimpleObject` only ever emits `results`, whereas `NeedsPaging` goes through `BasePageableResult`, which adds `totalCount` when the request asked for it.

Concretely: `GET /ws/rest/v1/encounter?patient=5946f880-b197-400b-9caa-a3c661d23041&fromdate=2099-01-01&totalCount=true` (a patient that exists, a date range that matches nothing) returns `{"results": [], "totalCount": 0}` on 4.x and `{"results": []}` on this branch. So a client using `totalCount` to drive its pager can no longer tell "zero rows" from "the server didn't answer the question", and since nothing throws, no test or CI run will catch it.

```suggestion
	private PageableResult page(List<D> results, RequestContext context) {
		return new NeedsPaging<D>(results, context);
	}
```

That leaves the `EmptySearchResult` import unused. I ran the four existing `EncounterSearchHandler2_0Test` cases with this applied and they still pass, with `totalCount: 0` back on the empty search.

One thing the empty branch was incidentally covering up: `NeedsPaging.getPageOfResults()` clamps `endIndex` to the list size but not `startIndex`, so an empty list plus `startIndex=5` throws `IllegalArgumentException: fromIndex(5) > toIndex(0)`. That already happens on 4.x, and on this branch it still happens for a non-empty overshoot (3 results, `startIndex=5`), so it looks better fixed in `NeedsPaging` than half-hidden here.

There is also no test over `TypedSearchHandler` itself yet, and the zero-results-plus-`totalCount` case is the one I would add first.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchConfig.java
```diff
@@ -0,0 +1,90 @@
...
+      .withOptionalParameters(optionalParams.toArray(new String[optionalParams.size()]))
+      .build();
+  }
+
+  /**
+	 * Get this {@code searchQueries}.
+	 * 
+	 * @return this search queries
+	 */
+	public SearchQuery getSearchQuery() {
```
dkayiwa, line 79: This file is indented with 2 spaces where the rest of omod-common uses tabs, and it mixes the two (this method is tab indented, the 40-odd lines around it are not). Worth reformatting by hand, because the java-formatter plugin in the build isn't actually formatting anything at the moment: I mis-indented a line in `SearchQuery.java`, ran it, and the mis-indentation survived.

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/search/openmrs2_0/EncounterSearchHandler2_0.java
```diff
@@ -22,49 +22,60 @@
...
-			Collections.singletonList("2.0.* - 9.*"),
-			Collections.singletonList(new SearchQuery.Builder(
-					"Allows you to find Encounter by patient and encounterType (and optionally by from and to date range)")
-					.withRequiredParameters("patient").withOptionalParameters("visit", "encounterType", DATE_FROM, DATE_TO,
-							"order", "totalCount").build()));
-	
-	@Override
-	public SearchConfig getSearchConfig() {
-		return this.searchConfig;
+public class EncounterSearchHandler2_0 extends TypedSearchHandler<Encounter, EncounterResource2_2, EncounterSearchHandler2_0.Params> {
```
mogoodrich, line 38: this search handler supported versions of OpenMRS from 2.0 and above... but now it is being changed to use EncounterResource, so assumedly it may no longer support 2.0 and 2.1 anymore?  (At the end of the day we don't want to worry about going through hoops to support super-old versions, but wanted to make sure we've thought about this)

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/resource/api/TypedSearchHandler.java
mogoodrich: We might want to change this at some point (but I agree with maintaining the default for now)

mogoodrich (changes requested):
Generally trust you @chibongho but don't have a change to go through with a fine-toothed comb.

Can we make sure we have pretty extensive search coverage on the EncounterSearchHandler class?  And maybe there'd be some ways to test abstract class itself?

PR #765, opened 2026-09-16
RESTWS-1051: Add effectiveReferenceRange property to Obs resource

mseaton:
## Summary
- Adds an `effectiveReferenceRange` property to `ObsResource2_7` (opt-in via custom representation, not included in the DEFAULT representation): the reference range directly associated with the obs if there is one, otherwise the reference range associated with its concept, evaluated as of the obs's own date via `ConceptReferenceRangeContext(obs)` so date-relative criteria (e.g. age-at-encounter) are evaluated correctly for historical results rather than as of today.
- Unlike `referenceRange` (which only reflects a direct obs-level association and is left unchanged), `effectiveReferenceRange` is always populated from either source in a single consistent shape (`hiNormal`, `hiAbsolute`, `hiCritical`, `lowNormal`, `lowAbsolute`, `lowCritical`), so a client doesn't need to know or care which source it came from.
- Originated from a PIH pihapps module use case (lab result display falling back to concept-level reference ranges); this PR moves that logic into core so it's broadly available to any consumer of `/ws/rest/v1/obs`, rather than maintained as a custom downstream resource.

## Test plan
- [x] `mvn compile` / `mvn test-compile` pass
- [x] Added `ObsController2_7Test` covering: direct range mirrored, concept-level fallback when no direct range exists, and absence when not requested
- [ ] Could not execute the new integration test in this environment (missing `omod/src/main/webapp/WEB-INF/view/module/legacyui/` at test runtime, a pre-existing local checkout gap unrelated to this change — reproduced identically against the existing, unmodified `ObsReferenceRangeController2_7Test`). Please run in CI/a fully-bootstrapped environment before merging.

This targets 4.x per plan to up-port to master once tested and approved.


omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_7/ObsResource2_7.java
```diff
@@ -30,7 +35,35 @@ public DelegatingResourceDescription getRepresentationDescription(Representation
...
+	 * age-at-encounter are evaluated correctly for historical results, not as of today). Unlike
+	 * "referenceRange", which only reflects a direct obs-level association, this is populated from either
+	 * source in a single consistent shape, so a client does not need to know or care which source it came
+	 * from.
+	 */
+	@PropertyGetter("effectiveReferenceRange")
+	public SimpleObject getEffectiveReferenceRange(Obs obs) {
+		BaseReferenceRange effectiveRange = obs.getReferenceRange();
+		if (effectiveRange == null) {
+			effectiveRange = Context.getConceptService().getConceptReferenceRange(new ConceptReferenceRangeContext(obs));
```
dkayiwa, line 51: This getter cannot run on most of the platforms this resource is registered for. `ConceptReferenceRangeContext` and the `getConceptReferenceRange(ConceptReferenceRangeContext)` overload first appear in openmrs-api 2.8.5, but the `@Resource` annotation above still says `2.7.* - 9.*`, so on 2.7.0 through 2.8.4 this is the class serving `/obs`, and any request for `v=custom:(effectiveReferenceRange)` fails there with `NoClassDefFoundError: org/openmrs/ConceptReferenceRangeContext` instead of returning a range.

I checked every released jar rather than trusting the `@since` tag: none of 2.7.0 to 2.7.9 has the class (it is not on core's 2.7.x branch at all, so the 2.7.9 backport the javadoc claims never shipped), none of 2.8.0 to 2.8.4 has it, and 2.8.5 onwards do. Running this branch's compiled `ObsResource2_7.class` with openmrs-api 2.8.3 and 2.7.8 on the classpath reproduces the error at the `new`. The class itself still loads and instantiates, and `findPropertyGetterMethod` only looks this up by name, so the breakage is confined to requests that actually ask for the property.

The module already solves this one package over, for this very API: `ConceptReferenceRangeResource2_7` is pinned to `"2.7.0 - 2.7.8, 2.8.0 - 2.8.4"` and `ConceptReferenceRangeResource2_8` carries the context-based code. An `ObsResource2_8 extends ObsResource2_7` under `resource.openmrs2_8` holding this getter, with `ObsResource2_7` narrowed to the older range, would work the same way here. Two things to watch if you go that way: the two ranges have to be disjoint, or `RestServiceImpl.isResourceToBeAdded` throws "Two resources with the same name must not have the same order" at startup; and I would not copy the `2.7.9 - 2.7.999` half of `ConceptReferenceRangeResource2_8`'s range, since released 2.7.9 does not have the class either (that looks like a separate, pre-existing problem).

omod/src/test/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs2_7/ObsController2_7Test.java
```diff
@@ -0,0 +1,79 @@
...
+		SimpleObject response = deserialize(handle(newGetRequest("obs/2f616900-5e7c-4667-9a7f-dcb260abf1de",
+		    new org.openmrs.module.webservices.rest.web.v1_0.controller.MainResourceControllerTest.Parameter("v", "custom:(effectiveReferenceRange)"))));
+
+		Object effectiveReferenceRange = PropertyUtils.getProperty(response, "effectiveReferenceRange");
+		Assert.assertEquals(300.0, PropertyUtils.getProperty(effectiveReferenceRange, "hiAbsolute"));
+		Assert.assertEquals(350.0, PropertyUtils.getProperty(effectiveReferenceRange, "hiCritical"));
+	}
+
+	@Test
+	public void shouldFallBackToConceptLevelReferenceRangeWhenObsHasNoDirectReferenceRange() throws Exception {
```
dkayiwa, line 61: Neither of these tests pins the behavior that motivates the change. I swapped the resource's call for the pre-2.8.5 `getConceptReferenceRange(obs.getPerson(), obs.getConcept())`, which evaluates criteria as of today rather than as of the obs date, and all 7 tests still passed. Nothing exercises a matching `concept_reference_range` row either: this case lands on the `concept_numeric` defaults, which is worth reflecting in the method name, since `shouldFallBackToConceptLevelReferenceRange` reads like the `concept_reference_range` path.

Here is a case that does discriminate, which I ran against this branch. Add a dataset row for concept 5089 with `criteria="$patient.getAge($date) &lt;= 40"` and `hi_normal="99.0"`, then GET the same obs: person 7 was born 1976-08-25 and the obs is dated 2008-07-01, so she is 31 as of the obs and 50 today. On this branch `hiNormal` comes back as 99.0; with the person/concept call it falls through to the 250.0 default.

omod/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/resource/openmrs2_7/ObsResource2_7.java
```diff
@@ -21,7 +26,7 @@
  */
 @Resource(name = RestConstants.VERSION_1 + "/obs", supportedClass = Obs.class, supportedOpenmrsVersions = { "2.7.* - 9.*" })
 public class ObsResource2_7 extends ObsResource2_1 {
-	
+
```
dkayiwa, line 29: This blank line and the one before the new javadoc lost their leading tab, which is not part of the change. The other files in this package still have them, so could you put them back and keep the diff to just the new property?

PR #766, opened 2026-09-16
RESTWS-1052: Apply order arbitration to class-based resource lookup

mseaton:
## Summary
`RestServiceImpl.initializeResources()` builds two lookup maps: `resourceDefinitionsByNames` (order-arbitrated on name collisions) and `resourcesBySupportedClasses` (no arbitration at all — last one scanned wins, regardless of `order`).

Because `RestService.getResourceBySupportedClass(Class)` and `ConversionUtil.getConverter(Class)` (which delegates to it) both read from `resourcesBySupportedClasses`, a resource can be silently and non-deterministically replaced by an unrelated, differently-named resource from another module that happens to support the same domain class — even when the replaced resource has a better `order` and would win a same-name comparison. Core's own `MainResourceController.retrieve()` resolves by name, so it's unaffected; custom code that resolves by class is exposed.

This fix applies the same `order`-based arbitration already used for `resourceDefinitionsByNames` to `resourcesBySupportedClasses`.

Fixes RESTWS-1052.

## Test plan
- Added `getResourceBySupportedClass_shouldNotBeClobberedByADifferentlyNamedResourceWithAWorseOrder` and `getResourceByName_shouldReturnResourceThatLostTheBySupportedClassLookup` to `RestServiceImplTest`, using a new fixture `AnimalWithHigherOrderResource_1_9`.
- Verified manually (via `git stash`) that the first new test fails without this fix and passes with it.
- Full `omod` test suite passes (55/55 in `RestServiceImplTest`).


omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -160,35 +160,67 @@ private void initializeResources() {
...
 			ResourceMetadata resourceMetadata = getResourceMetadata(resource);
 			if (resourceMetadata == null)
 				continue;
-			
-			if (isResourceToBeAdded(resourceMetadata, tempResourceDefinitionsByNames.get(resourceMetadata.getName()))) {
+
+			boolean preferredForName = isResourceToBeAdded(resourceMetadata,
+			    tempResourceDefinitionsByNames.get(resourceMetadata.getName()));
+			boolean preferredForClass = isResourceOrderBetter(resourceMetadata.getOrder(),
+			    tempOrdersBySupportedClasses.get(resourceMetadata.getSupportedClass()));
```
dkayiwa, line 184: This is the part I think needs settling before the PR goes in. `order` defaults to `Integer.MAX_VALUE`, so as soon as it is compared across resources that do not share a name, any resource that sets an order at all outranks every resource that leaves it alone, and almost nothing sets one: 124 of the 143 annotated resource classes in this repo leave it at the default.

The casualty is the pair from the ticket. `ObsResource2_1` and `ObsResource2_7` declare no `order`, coreapps' `LatestObsResource` declares `order = 1`, so on any 2.1+ platform with coreapps installed `Obs.class` resolves to `LatestObsResource` whatever the scan order is. I fed the real `ObsResource2_7` plus a class shaped exactly like `LatestObsResource` (name `v1/latestobs`, `order = 1`, `supportedClass = Obs.class`) to `RestServiceImpl` in both scan orders: on 3.x the winner flips with the order, on this branch it is `LatestObsResource` both times, while `getResourceByName("v1/obs")` still returns `ObsResource2_7`. `LatestObsResource extends ObsResource1_11`, so anything resolved by class rather than by name, which is every obs nested in an encounter or a visit because `DelegatingResourceDescription.Property.evaluate` goes through `ConversionUtil.getConverter`, comes back in the 1.11 representation with no `status`, `interpretation` or `referenceRange`, and with a `self` link of `/ws/rest/v1/latestobs/<uuid>`. Today that is a coin flip per deployment. After this it is certain. The queue module is the same shape, with `QueueEntrySubResource` (`order = 10`) taking `QueueEntry.class` off `QueueEntryResource` (no order).

`order` is only defined relative to a name. That is what `isResourceToBeAdded`'s own "Two resources with the same name (...) must not have the same order" is saying, and it is why the default is MAX rather than 0. Keeping the comparison inside that scope would work: store the owning name next to the order, and let a candidate replace the incumbent only when the names match and the order is better. That still fixes what you hit, since `ExtendedObsResource` and core's obs resource are both `v1/obs`, and it cannot hand `Obs.class` to `v1/latestobs`.

What is left over is the genuine cross-name collision, and `order` cannot settle that one either: the usual case is two defaults, which `candidateOrder < existingOrder` gives to whoever was scanned first. `v1/conceptsearch` against `v1/orderable`, `v1/concepttree` against `v1/obstree`, and bahmni's `v1/bahmniprogramenrollment` against `v1/programenrollment` are all in that state today and stay in it after this change, which is the half of the report about getting one answer in your SDK and another on the test server. Rather than pick one silently I would log a warning naming both resources and the class, since the silence is what makes this so expensive to track down.

Smaller point on the tests: both new cases give the incumbent `order = 1` and the challenger `order = 5`. One that leaves the incumbent at the default order is what would have caught the above.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
```diff
@@ -160,35 +160,67 @@ private void initializeResources() {
 		if (resourceDefinitionsByNames != null) {
 			return;
 		}
-		
+
```
dkayiwa: This blank line, and the ones at 175 and 205 (plus one in `RestServiceImplTest` just above the new tests), lost their indentation, which is nothing to do with the fix. Putting them back keeps the diff to the arbitration change.

omod-common/src/main/java/org/openmrs/module/webservices/rest/web/api/impl/RestServiceImpl.java
ibacher: Generally, I think treating entities with no defined order as lower-precedence than entities with a defined order is the right behavior. We want "core" modules to be implementation overridable and that's a straight-forward way to achieve that.

I think the issue here is that the `/latestobs` resources, i.e., maybe it should be a plain controller rather than a resource.

mseaton: I am wary about breaking backwards compatibility with things that have been working in a certain way.  Ultimately, this PR came out of the fact that I was trying to do this:

`Converter<Obs> obsConverter = ConversionUtil.getConverter(Obs.class);`

I wasn't actually trying to query for a resource at all, I was trying to query for a converter.  It's just that generally the resources are also converters and our logic for retrieving them goes through that mechanism.

ibacher (approved):

